PR #10, opened 2014-06-12
Update boostrap and list of all available responses

SiggyF:
I updated the bootstrap.py script to the latest version. It gave me some problems with setuptools/distutils.
The source is http://downloads.buildout.org/2/bootstrap.py as documented in https://pypi.python.org/pypi/zc.buildout/2.2.1. 

The other thing I changed is the list of responses in the index.html. I added the full list of available responses to the index page. 

The gsw package is listed as a dependency and gsw is dependent on scipy (loadmat). The gsw is missing scipy in the setup.py. So I had to (temporary) add scipy as a dependency to setup.py and the blas/atlas package to the travis build script, otherwise scipy won't be found. 
~~The travis build is now quite slow. You can use the version that pacificclimate made:
https://github.com/pacificclimate/pydap/commit/707b15b855c6f662deb927c2fe0383ab4661b662
That drops build support for python2.6.~~
I had to use the system scipy. It does not build under travis. Because there is no system version of scipy for python 2.6 I had to disable 2.6 in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ python:
   - "2.7"
 install:
   - "pip install -e .[testing] --use-mirrors"
+before_install:
+  - sudo apt-get update -qq
+  - sudo apt-get install -qq libatlas-dev libatlas-base-dev liblapack-dev gfortran
+
 script: nosetests
 branches:
   only:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 language: python
 python:
-  - "2.6"
   - "2.7"
 install:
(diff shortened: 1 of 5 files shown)


laliberte:
@SiggyF @jameshiebert do we have an idea which problem this PR is attempting to solve?

jameshiebert:
I think that this PR is quite out of date. We don't use bootstrap anymore (I don't think), we've dropped support for py26 and the gsw/scipy deps have been sorted out. And parts of this PR (the html template, I think) were submitted as another PR IIRC. I think that it's safe to close this w/o merging.

PR #13, opened 2015-09-27
fixed use of deprecated string.strip

jfburkhart:
I was having some problems with the use of the string module in Python 3.3. I fixed it with a lambda.


jameshiebert:
Thanks for the contribution, however this will be supplanted in the develop branch once the new Python 3 support gets merged in #9.

PR #21, opened 2016-08-02
Show all BaseType variable's dimensions in the HTML form

ajelenak-thg:
The HTML form now displays all the dimensions of a BaseType variable and not only the first one as currently.


ajelenak-thg:
Made a mistake, wrong payday branch. Will close this PR and create a new one.

PR #22, opened 2016-08-02
Show all BaseType variable's dimensions in the HTML form

ajelenak-thg:
The HTML form now displays all the dimensions of a BaseType variable and not only the first one as currently.


jameshiebert:
Thanks @ajelenak-thg! I've merged this into develop (with an additional commit that ensures that all the tests pass). Cheers!

Issue #23, opened 2016-09-15
AttributeError: module 'string' has no attribute 'strip' when installing with Python3


lewismc:
Hi folks, [as noted some time ago](https://groups.google.com/a/opendap.org/forum/#!msg/support/-_15rXI-4dI/vrd0_GvjCwAJ) Pydap cannot be installed using Python3.
This issue should fix the issue below.

```
Collecting pydap>=3.1.1 (from -r ocw-pip-dependencies.txt (line 3))
  Using cached Pydap-3.1.1.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-bpljh_u8/pydap/setup.py", line 10, in <module>
        paver.tasks.main()
      File "paver-minilib.zip/paver/tasks.py", line 883, in main
      File "paver-minilib.zip/paver/tasks.py", line 851, in _launch_pavement
      File "pavement.py", line 31, in <module>
        classifiers=filter(None, map(string.strip, '''
    AttributeError: module 'string' has no attribute 'strip'
```

lewismc:
This PR I am sending also ensures that all of the code is PEP8 compliant and upgrades to Python 3 syntax.

lewismc:
Thiis issue has been addressed within issue #9

PR #24, opened 2016-09-15
AttributeError: module 'string' has no attribute 'strip' when installing with Python3

lewismc:
Hi Folks, this issues addresses #23 @robertodealmeida @SiggyF @shoyer

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2016 Pydap
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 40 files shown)


shoyer:
I appreciated your interest here, but I believe @jameshiebert has been working on Python 3 support over in #9. I don't know what the current status of that PR is, though.

jameshiebert:
@shoyer is right. I have Python 3 support at #9 that I think is pretty close to ready to go, but it has been ignored for waaaaay too long, while I've been pulled in many directions. @lewismc if you want to get involved in this, would you mind checking out that branch testing out whether it meets your needs?

lewismc:
@jameshiebert yes I definitely will. Thanks. My main issue is that I can't use the existing most recent pydap artifact in a Python 3 build at all. I'll close this issue off and we can work over on #9 . Thanks

Issue #20, opened 2016-07-28
Merge URS authentication branch


jgallagher59701:
The Merge request for the URS authentication seems like it's languishing. See Pull request #11 (Added support for NASA/URS SSO service)

Thanks,
James

bucricket:
Thanks to James for adding this.  This is really needed now that NASA requires logins through https://urs.earthdata.nasa.gov.  Please merge this!

Thanks!
Mitch

pmacaodh:
I second that and would rather use the production version than the branch to access the data.  Thanks.

jameshiebert:
Addressed by #28

PR #26, opened 2016-11-19
Added support for NASA/URS SSO service

lewismc:
@jgallagher59701 this is an attempt which builds off of your patch and also enables the test to run as part of the unit test invocation. ```nosetests``` will do it!
Can you please provide a .nc file present somewhere that requires URS authentication so we can test this? Thanks

### DIFF
--- a/src/pydap/net.py
+++ b/src/pydap/net.py
@@ -3,7 +3,12 @@ from webob.exc import HTTPError
 
 from six.moves.urllib.parse import urlsplit, urlunsplit
 
-def GET(url, application=None):
+# Should HTTPS require a valid certificate? Setting this to True will cause
+# sites with self-signed certificates to fail; False will enable HTTPS to 
+#  work with those sites. jhrg 4/22/15
+SSL_VALIDATE = True
+
+def GET(url, application=None, ssl_validate=True):
     """Open a remote URL returning a webob.response.Response object
 
     Optionally open a URL to a local WSGI application
--- a/src/pydap/responses/urs.py
+++ b/src/pydap/responses/urs.py
@@ -1,0 +1,105 @@
+"""
+LGPL License
+Copyright (c) 2015
+All rights reserved.
+
+PyDAP extension to enable access to OpenDAP servers that 
+require login using NASA URS. This code will actually provide
+credentials to any server that uses a 401 response to ask for
+them. Of course, HTTP Basic authentication sends the information
+in the clear, so HTTPS should always be used by the server.
+
+@author: James Gallagher <jgallagher@opendap.org>
+
+"""
+
+from pydap.net import GET
+from six.moves.http_cookiejar import CookieJar
+from six.moves.urllib.request import build_opener
+from six.moves.urllib.request import HTTPBasicAuthHandler
+from six.moves.urllib.request import HTTPCookieProcessor
+from six.moves.urllib.request import HTTPPasswordMgrWithDefaultRealm
+from six.moves.urllib.request import install_opener
+from six.moves.urllib.request import urlopen
+import netrc
+import re
+
+import pydap.lib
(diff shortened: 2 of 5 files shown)


src/pydap/tests/test_urs.py
```diff
@@ -0,0 +1,17 @@
...
+import logging
+
+logging.basicConfig(filename='urs_test.log',level=logging.DEBUG)
+
+class TestUrs(unittest.TestCase):
+
+    def test_basic_urs_auth(self):
+        """Set up PyDAP to use the URS request() function"""
+        install_basic_client('https://urs.earthdata.nasa.gov','pydap', 'OPeNDAP1', False)
+        self.assertIsNotNone(open_url('https://test.opendap.org/opendap/data/nc/coads_climatology.nc'))
```
lewismc: @jgallagher59701 does this file require URS authentication in order to access?

jameshiebert:
Awesome! I'd like to test this out (and my attempt at the same thing), but I've had trouble getting my NASA/URS account request through. The validation e-mail never arrives (I've made a couple attempts in the last month). Has anyone else had problems with this or know how I can get my account validated?

lewismc:
@jameshiebert 

In our tests, I've used a generic login username and password, you will see if you look into the test case code.

> Has anyone else had problems with this or know how I can get my account validated?

No I have never had problems however I can point you where you can resolve them. Can you please email support@earthdata.nasa.gov with your query... your PoC there may end up being Nemanja Djuric.
Also please visit the URL below and request an account to be set up.
https://support.earthdata.nasa.gov/index.php?
hth

src/pydap/net.py
```diff
@@ -3,7 +3,12 @@
 
 from six.moves.urllib.parse import urlsplit, urlunsplit
 
-def GET(url, application=None):
+# Should HTTPS require a valid certificate? Setting this to True will cause
+# sites with self-signed certificates to fail; False will enable HTTPS to 
+#  work with those sites. jhrg 4/22/15
+SSL_VALIDATE = True
```
lewismc: Currently this is not used anywhere.

src/pydap/responses/urs.py
```diff
@@ -0,0 +1,105 @@
...
+        r = urlopen(url)
+        
+        resp = dict([i[0], i[1]] for i in r.headers.items())
+        resp['status'] = str(r.code)
+        data = r.read()
+
+        # When an error is returned, we parse the error message from the
+        # server and return it in a ``ClientError`` exception.
+        if resp.get("content-description") == "dods_error":
+            m = re.search('code = (?P<code>\d+);\s*message = "(?P<msg>.*)"',
```
lewismc: I've never managed to step into this code yet either... so it may throw an Exception as I don't believe that 're' is instantiated before we try to use it.

jameshiebert:
> Can you please email support@earthdata.nasa.gov with your query.

Sure, and done. I'll see how that goes.

src/pydap/net.py
jameshiebert: Agreed. Strike it.

src/pydap/responses/urs.py
jameshiebert: `re` *is* imported above (line 57). But yeah, it'd be ideal to get test coverage across these lines. That's one reason to use `pydap.client.open_url()`, because you can actually pass a WSGI app as a second arg to the function. Then in the tests, you can write a fake WSGI app that just returns a dods_error, and use it to test this section of code. If you look in `src/pydap/tests/test_client.py` you can see multiple places where this strategy is employed.

jameshiebert:
This functionality *should* be addressed by #20 which has been merged into master. If it's not for some reason, feel free to re-open this PR.

PR #29, opened 2016-12-03
Fixing test_cas_urs.py

laliberte:
I've been having trouble testing the URS tests with the ``assert session.auth`` check. I've rebased to the latest master and I'm still having trouble. See my travis-ci with auth tests: [auth testing latest commits](https://travis-ci.org/laliberte/pydap/builds/180879857). This PR removes the ``assert session.auth``. This allows one to succesfully test the URS authentication.

@jameshiebert, it might be a dumb question but why do you think it important to include ``assert session.auth``?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ install:
     - pip install -e .[testing,functions,tests]
 
 script:
-    - python setup.py nosetests
+    - python setup.py nosetests -a '!auth'
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,4 +17,4 @@ install:
     - pip install -e .[testing,functions,tests]
 
 script:
-    - python setup.py nosetests -a '!auth'
+    - python setup.py nosetests
--- a/src/pydap/tests/test_cas_urs.py
+++ b/src/pydap/tests/test_cas_urs.py
@@ -42,7 +42,7 @@ class TestUrs(unittest.TestCase):
         session = urs.setup_session(os.environ.get('USERNAME_URS'),
                                     os.environ.get('PASSWORD_URS'),
                                     check_url=self.url)
-        assert session.auth
+
         dataset = open_url(self.url, session=session)
         expected_data = [[[99066.15625, 99066.15625, 99066.15625,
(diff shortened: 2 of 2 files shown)


jameshiebert:
> it might be a dumb question but why do you think it important to include assert session.auth?

No, that's not a dumb question. Thanks for asking to clarify.

My reasoning was that if a developer is running the test suite, but doesn't have credentials set, that test is going to *error*. And I would argue that tests should *never* error... they should either pass or fail. So trace the flow of the test [here](https://github.com/pydap/pydap/blob/master/src/pydap/tests/test_cas_urs.py#L42) supposing that the auth env variables are not set:

```python
    session = urs.setup_session(os.environ.get('USERNAME_URS'),
                                    os.environ.get('PASSWORD_URS'),
                                    check_url=self.url)
    assert session.auth
    dataset = open_url(self.url, session=session)
```

On the first line, setup_session() is called with blank strings (or None... not sure which). Either way, it handles the empty credentials fine and returns a session with cookies. However, the requests that it makes return 401s, so while the session gets returned, ``session.auth`` will be None (I think).

Without checking whether that exists (i.e. if you don't ``assert session.auth`` first), ``open_url`` will send a request, and call ``raise_for_status`` for anything other than a 200 OK. So if the response comes back as another 401, this test will raise an exception and have an error status (rather than the desirable fail status).

I *thought* that simply asserting for truthiness on ``session.auth`` would work, but I *still* don't have my credentials from URS (after asking through three different venues over the course of the last 6 weeks... grrrrrrr). If that doesn't work for you, the other way to achieve the same thing would be to catch a very specific exception (catch the HTTPError and then check the status code) and fail the test if that happens. Something like this:

```python
try:
    dataset = open_url(self.url, session=session)
except HTTPError as e:
    if e.detail == 401:
        # Fail the test some how. assert False?
    else:
        raise e
```

Does that seem doable?

laliberte:
Ok, I got it! Makes complete sense. I'm going to withdraw this PR and resubmit another once testing is over. I'm proposing to simply use the session with a test url. I works and it's feels pretty clean. You'll be the judge.

PR #31, opened 2016-12-03
Fix a bug with ESGF OPENDAP servers

laliberte:
This PR solves a bug on ESGF servers whereas retrieving coordinates of grid variables would  uncover a memory leak on the OPENDAP server. The solution proposed here is to provide a boolean option ``open_url(..., output_grid=False)`` that disables the retrieval of these coordinates. This essentially reverts to the behavior of ``pydap 3.1.1`` and is more akin to the behavior of ``netcdf4-python``. 

Considering that many use cases for ``pydap`` involve the ESGF, I would consider this bug fix high priority (it certainly is for me: without it ``pydap 3.2.0`` is essentially useless for me). I am open to a different approach / another name for ``output_grid``, as long as there is a high level method to disable the grid retrieval.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -147,29 +147,29 @@ class BaseProxy(object):
             int(np.ceil((s.stop-s.start)/float(s.step))) for s in index)
         size = int(np.prod(shape))
 
-        try:
-            if self.dtype == np.byte:
-                return np.fromstring(data[:size], 'B')
-            elif self.dtype.char in 'SU':
-                out = []
-                for word in range(size):
-                    n = np.fromstring(data[:4], '>I')  # read length
-                    data = data[4:]
-                    out.append(data[:n])
-                    data = data[n + (-n % 4):]
-                return np.array([ text_type(x.decode('ascii')) for x in out ], 'S')
-            else:
+        if self.dtype == np.byte:
+            return np.fromstring(data[:size], 'B')
+        elif self.dtype.char in 'SU':
+            out = []
+            for word in range(size):
+                n = np.fromstring(data[:4], '>I')  # read length
+                data = data[4:]
+                out.append(data[:n])
+                data = data[n + (-n % 4):]
+            return np.array([ text_type(x.decode('ascii')) for x in out ], 'S')
+        else:
+            try:
                 return np.fromstring(data, self.dtype).reshape(shape)
-        except ValueError as e:
-            if str(e) == 'total size of new array must be unchanged':
-                # server-side failure. do not fail. instead, return NaNs
-                # it is expected that the user should be mindful of this:
-                raise RuntimeError('varirable {0} could not be properly '.format(quote(self.id)) +
-                                   'retrieved. '
-                                   'To avoid this '
-                                   'error consider using open_url(..., output_grid=False).')
-            else:
-                raise
+            except ValueError as e:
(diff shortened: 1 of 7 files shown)


laliberte:
The initial commit [tests ok with my credentials](https://travis-ci.org/laliberte/pydap/builds/180898279). I'm removing auth tests and push a new version.

laliberte:
I've been thinking about this one and another option would be attempt to mimic the grid type by retrieving the variables from the parent dataset. It would however require to implement some sort of parent tracking mechanism in the objects defined in ``models.py``. It would be some work but it could potentially be useful for other usages. And we could drop the extra keyword, etc.

src/pydap/model.py
```diff
@@ -564,6 +567,16 @@ def __getitem__(self, key):
             return out
 
     @property
+    def output_grid(self):
+        if hasattr(self, '_output_grid'):
```
jameshiebert: It might simplify the code to ``set _output_grid()`` as a default value (e.g. True) in ``__init__()``. Then you don't have to check whether it exists... it will be safe to assume that it does.

src/pydap/tests/test_cas_esgf.py
```diff
@@ -67,13 +67,15 @@ def test_dimension_esgf_query(self):
...
         assert(np.isclose(data, expected_data).all())
 
-    @unittest.skip("This test should work but does not. "
-                   "An issue should be raised.")
     def test_variable_esgf_query(self):
         session = esgf.setup_session(os.environ.get('OPENID_ESGF'),
                                      os.environ.get('PASSWORD_ESGF'),
                                      check_url=self.url)
-        dataset = open_url(self.url, session=session)
+        # This server does not access retrieval of grid coordinates.
```
jameshiebert: This comment is nice and descriptive, but could you write up a bit of this information in the user docs? It would be helpful for users to know about this behaviour w/o having to read the code.

src/pydap/handlers/dap.py
```diff
@@ -143,18 +147,29 @@ def __getitem__(self, index):
...
+                out = []
+                for word in range(size):
+                    n = np.fromstring(data[:4], '>I')  # read length
+                    data = data[4:]
+                    out.append(data[:n])
+                    data = data[n + (-n % 4):]
+                return np.array([ text_type(x.decode('ascii')) for x in out ], 'S')
+            else:
+                return np.fromstring(data, self.dtype).reshape(shape)
+        except ValueError as e:
```
jameshiebert: This is pretty coarse exception handling. Which part of the block above actually raises the `ValueError`?

src/pydap/model.py
laliberte: I've added an ``__init()__`` to gridtypes. It is indeed cleaner.

src/pydap/tests/test_cas_esgf.py
laliberte: I've added a new subsection in the client docs explaining this option.

src/pydap/handlers/dap.py
laliberte: I've moved it closer to the reshape, which is the operation causing the error. I don't know why I've put it so far away. Thanks for pointing this out.

jameshiebert:
Looks good! Thanks for addressing those changes; I'll go ahead and merge now.

PR #32, opened 2016-12-03
Some cleaning and flake8 checking

laliberte:
Here I am proposing a simple (yet extensive) clean-up of the latest code. I have run ``flake8`` on all files and I have included it in the python 3.5 testing branch.

This is just a proposal but I have found that systematically using flake8 makes it easier to maintain a uniform-ish code. 

Is this something you guys would think desirable for future ``pydap`` developments?

### DIFF
--- a/src/pydap/tests/test_handlers_lib.py
+++ b/src/pydap/tests/test_handlers_lib.py
@@ -480,10 +480,9 @@ class TestNestedIterData(unittest.TestCase):
     def test_filter_data(self):
         """Test filtering the data."""
         self.assertEqual(list(self.data[self.data["lat"] > 2]),
-                         [(3, 6, 9, []), (4, 8, 16, [(31, 32, 33),
-                                                     (41, 42, 43),
-                                                     (51, 52, 53),
-                                                     (61, 62, 63)])])
+                         [(3, 6, 9, []), (4, 8, 16,
+                                          [(31, 32, 33), (41, 42, 43),
+                                           (51, 52, 53), (61, 62, 63)])])
 
     def test_deep_filter(self):
         """Test deep filtering the data."""
--- a/src/pydap/tests/test_handlers_lib.py
+++ b/src/pydap/tests/test_handlers_lib.py
@@ -2,14 +2,8 @@
 
 import sys
 import copy
-import re
-if sys.version_info < (2, 7):
-    import unittest2 as unittest
-else:
-    import unittest
 from six import text_type
 
-from pkg_resources import EntryPoint
 from webtest import TestApp, AppError
 import numpy as np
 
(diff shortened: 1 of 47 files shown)


src/pydap/tests/test_handlers_lib.py
```diff
@@ -417,96 +417,82 @@ def setUp(self):
...
         self.assertEqual(list(self.data["time_series"][1]), [[(15, 16, 17)]])
 
     def test_filter_data(self):
         """Test filtering the data."""
-        self.assertEqual(
-            list(self.data[self.data["lat"] > 2]), [
-                (3, 6, 9, []),                                                              
-                (4, 8, 16, [(31, 32, 33), (41, 42, 43), (51, 52, 53), (61, 62, 63)]),
-            ])
+        self.assertEqual(list(self.data[self.data["lat"] > 2]),
```
jameshiebert: I don't think flake8 did a good job w/ this line. You might want to manually fix the spacing.

src/pydap/tests/test_responses_ascii.py
```diff
@@ -53,24 +51,25 @@ def test_headers(self):
...
-        Int32 pressure;
-    } cast;
-} SimpleSequence;
----------------------------------------------
-cast.id, cast.lon, cast.lat, cast.depth, cast.time, cast.temperature, cast.salinity, cast.pressure
-"1", 100, -10, 0, -1, 21, 35, 0
-"2", 200, 10, 500, 1, 15, 35, 100
-
-""")
+        self.assertEqual(self.res.text,
```
jameshiebert: Not sure that I like this better. IMO, this is much less readable than using the multiline string. Can we turn off flake8 testing for this (and the several cases like this that follow)?

laliberte:
I don't mind the rebasing, I've made sure all the PRs were kind of compatible by successively merging them. Don't worry about that ;)

src/pydap/tests/test_handlers_lib.py
laliberte: Some of it is my editorial decision... I'll take another jab at it!

src/pydap/tests/test_responses_ascii.py
laliberte: You are right that it is much less readable. I guess my idea here was that since it is a unittest, readability is important but what is more important is exactitude. This way, one sees precisely what characters are expected and extraneous spaces at EOLs are not hidden. We can easily turn off flake8 to that whole file, if you prefer.

laliberte:
@jameshiebert do you want to attempt to close this one today? I can push my latest version. What do we decide about multiline strings? Readability vs exactitude?

src/pydap/tests/test_responses_ascii.py
jameshiebert: Yes, this bugs me to have the explicit EOLs and all of the extra padding at the beginning. Python's multiline strings exist for a reason :) I think that if you revert these comparisons back to checking against a multiline string (preferred), or just turn off flake8 checking on this file, then I'll be happy and we can merge this.

laliberte: Ok, I've excluded this file from flake8. There are however other tests with multiline strings that were modified. Do you want me to do the same?

laliberte:
Ok, I've reverted most (if not all) multiline strings to their previous state. It required me to specifically exclude two files from ``flake8`` checking. These are excluded in ``setup.cfg``.

PR #34, opened 2016-12-12
Merging CSV handler and devel test server

laliberte:
This PR converts merges the CSV handler (that was up to now a separate github project) into the main branch. It also provides a simple test suite to test whether ``pydap`` serves the data seamlessly over a server.

My goal here was to provide the simplest test bed possible for a handler / server / client unit test while keeping requirements to a minimum. 

Any thoughts @jameshiebert  @shoyer?


laliberte:
I've modified this code this PR this morning and have separated the devel server from the test suite. It is now part of the main ``pydap`` code.

src/pydap/simple_server.py
```diff
@@ -0,0 +1,92 @@
...
+    >>> server.connect()
+    >>> dataset = open_url("http://localhost:%s" % server.port)
+    >>> server.disconnect()
+
+    Or as an instance:
+    >>> with LocalTestServer("test.csv", handler=CSVHandler) as server:
+    >>>     dataset = open_url("http://localhost:%s" % server.port)
+    """
+
+    def __init__(self, file_name, port=8000,
```
shoyer: It is possible to make this launch on a randomly assigned free port instead, e.g., `port=0`? That would make this far more robust.

 I know `dask.distributed` uses something similar for its similar test utility.

src/pydap/simple_server.py
```diff
@@ -0,0 +1,92 @@
...
+        return self
+
+    def disconnect(self):
+        # Shutdown the server:
+        (Request
+         .blank("http://0.0.0.0:%s/shutdown" % self.port)
+         .get_response())
+        time.sleep(self._wait)
+        self.server_process.join()
+        del(self.server_process)
```
shoyer: `del` is a keyword, not a function, so you don't need parenthesis

laliberte: Fixed!

src/pydap/simple_server.py
laliberte: Ok, I've figured it out. It seems to work and, yes, that would be much more robust!

laliberte:
I think this PR would be ready to be merged. I've moved the ``werkzeug`` from a testing pre-requisite to a new ``server`` extra. Any comments @jameshiebert ?

laliberte:
@jameshiebert Any chance this and #34 could be merged into master?

laliberte:
@jameshiebert:
What should we do about this one? I think that for testing purposes (especially if we want to include a timeout feature #37), it would be important to have a basic test server AND a basic handler (here I propose CSV because it has no dependencies). I don't think this commits us to supporting more handlers, it just means that we are willing to provide some basic test capabilities.
What do you think? Are we ready to merge this one?

jameshiebert:
Now that the test dev server is merged in, could you rebase this branch to resolve the conflicts, and then we can include it?

laliberte:
@jameshiebert I think this PR is pretty much completed.

Tests revealed a leftover bug from #89 that is now fixed to my liking (by introducing the __shallowcopy__ method). It is not perfect but I intend to revisit this issue very soon when tackling #76.

laliberte:
My commits don't pass for python 2.7. I'll fix this tomorrow.

laliberte:
@jameshiebert I think this is done. You can go ahead merge if it is to your liking!

PR #35, opened 2016-12-12
Merging netCDF handlers

laliberte:
This PR merges the previously disjoint netCDF handler into the main distribution.

As in #34 , the main goal here to provide a handler so that one could implement more advanced unit tests that would cover a more extensive array of use cases. One of my goal here is to provide a good testing environment to develop a ``netCDF4``-like client API for pydap as in [netcdf4_pydap](github/laliberte/netcdf4_pydap).

Unlike in #34, this PR introduces a major dependency, namely ``netCDF4``. However, I feel that since it is an optional pre-requisite it might not cause too much trouble.

I hope this can be useful to someone other myself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ addons:
       - netcdf-bin
       - libnetcdf-dev
 install:
-    - export USE_NCONFIG=1
-    - pip install netCDF4
+    - nc-config --all
+    - export USE_NCONFIG=1; pip install netcdf4
     - travis_wait 30 pip install scipy
     - pip install -e .[testing,functions,tests]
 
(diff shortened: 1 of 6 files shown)


laliberte:
This test currently fails for python 2.6 because ``netcdf4-python`` has dropped support for 2.6. Would it be time for ``pydap`` to drop support as well?

laliberte:
I was able to have a successful build even with python 2.6 by pointing to version 1.2.5 of ``netcdf4-python``.

shoyer:
+1 for dropping Python 2.6 support. Even NumPy is dropping Python 2.6 in its next release.

I don't think it's a problem to introduce optional netcdf4/scipy dependencies.

PR #37, opened 2016-12-12
Add a timeout feature and fix simple_server

laliberte:
This PR builds on #34 and introduces a ``timeout`` option to ``open_url()`` and ``open_dods``. Adding a ``timeout`` is of utmost importance since ESGF servers regularly fail to respond, which could lead to a hung process if ``pydap`` is used in a larger code.

I think this PR requires a test server since it is not possible to test for timeout simply using the WSGI interface.

This PR modifies the documentation to reflect the new timeout procedure and removes the discussion on the cache as it is no longer valid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,5 @@ install:
     - pip install -e .[testing,functions,server,tests]
 
 script:
-    - python setup.py nosetests -a '!auth'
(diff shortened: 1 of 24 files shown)


laliberte:
This PR has been superseded by #94.

PR #38, opened 2016-12-12
A bug fix in fix_slices

laliberte:
This PR fixes a bug that arises when a ``pydap`` array is sliced and the ``slice.stop`` integer is larger than the dimension length.


jameshiebert:
Ah, thanks for the great catch! LGTM.

PR #40, opened 2016-12-14
mechanicalsoup browser cannot be closed

laliberte:
This PR solves a overlooked bug: the mechanical soup browser cannot be closed. This line should be removed.

I could not come up with a good test for the offending piece of code. Sorry!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,9 +12,26 @@ addons:
       - libblas-dev
       - liblapack-dev
       - cython
+      # The following as required for netcdf files:
+      - libhdf5-serial-dev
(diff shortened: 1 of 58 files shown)


laliberte:
The latest commit removes the `notify` concept altogether. In fact this error-checking feature has been superseded by the check ``raise_if_form_exists``.

The problem with these errors is that they are needed mostly in high-volume and concurrent interactions with the ESGF. In such a situation, bugs are very difficult to reproduce.

I wish we could come up with an adequate testing mechanism but at this point, I don't se how.

Issue #43, opened 2016-12-21
model.GridType differences from 3.1.1


MBARIMike:
GridType in Version 3.1.1 has __len__() method and other attributes:

    # These attributes all come from the array object.
    def __len__(self): return len(self.array)
    @property
    def type(self): return self.array.type
    def _get_shape(self): return self.array.shape
    def _set_shape(self, shape): self.array.shape = shape
    shape = property(_get_shape, _set_shape)

Should they be added to maintain compatibility?

MBARIMike:
Upon further inspection I see that the GridType returned is different. In Version 3.1.1 I get a dictionary:

```python
>>> from pydap.client import open_url
>>> ds = open_url('http://dods.mbari.org/opendap/data/auvctd/surveys/2010/netcdf/Dorado389_2010_300_00_300_00_decim.nc')
>>> ds['temperature']
{'temperature': <pydap.model.BaseType object at 0x7f18464a7650>, 'time': <pydap.model.BaseType object at 0x7f18464a7750>}
```

With Version 3.2.0 I get something different:

```python
>>> from pydap.client import open_url
>>> ds = open_url('http://dods.mbari.org/opendap/data/auvctd/surveys/2010/netcdf/Dorado389_2010_300_00_300_00_decim.nc')
>>> ds['temperature']
<GridType with array 'temperature' and maps 'time'>
```

Perhaps this is an improvement and I should change my code?

taeold:
This explains some of the compatibility issue I saw in pydata/xarray#1174 - namely, datatype returned by `pydap.client` in version 3.2 is dictionary-_like_ but not quite so, missing couple of functions like `__len__` and `iteritems/items`.

Does it make sense to implement these functions? From my understanding of the code, `GridType`/`StructureType` appeared well equipped to implement those functions (am I naive in thinking so?).

MBARIMike:
It wasn't too hard for me to [make a work-around](https://github.com/stoqs/stoqs/pull/452/commits/44f487e5a5ba35e948dc46c615992e087620a533) in my code, but it'd be nice to have the same behavior that we have 3.1.1.

jenfly:
[...]

PR #48, opened 2017-01-03
Add NumPy-like properties to GridType and length to StructureType

laliberte:
This adresses #43. Tests are included and everything appears to be working fine.

### DIFF
--- a/src/pydap/model.py
+++ b/src/pydap/model.py
@@ -118,7 +118,7 @@ import copy
 from six.moves import reduce, map
 from six import string_types, binary_type
 import numpy as np
-from collections import OrderedDict, MutableMapping
+from collections import OrderedDict, Mapping
 
 from .lib import quote, decode_np_strings
 
@@ -283,7 +283,7 @@ class BaseType(DapType):
     data = property(_get_data, _set_data)
 
 
-class StructureType(DapType, MutableMapping):
+class StructureType(DapType, Mapping):
     """A dict-like object holding other variables."""
 
     def __init__(self, name, attributes=None, **kwargs):
--- a/src/pydap/model.py
+++ b/src/pydap/model.py
@@ -118,7 +118,7 @@ import copy
 from six.moves import reduce, map
 from six import string_types, binary_type
 import numpy as np
-from collections import OrderedDict
+from collections import OrderedDict, MutableMapping
 
 from .lib import quote, decode_np_strings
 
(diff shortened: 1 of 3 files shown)


laliberte:
@jameshiebert do you give your thumbs up? This one seems pretty straightforward.

shoyer:
It's worth double checking, but I think this is all we need for xarray's pydap integration to work again.

laliberte:
@shoyer, I'm not going to close the issue right away. If it's not sufficient, then we'll look at what's required. +1 for merging?

src/pydap/model.py
```diff
@@ -333,6 +333,9 @@ def keys(self):
         """Method to emulate a dictionary, returning keys."""
         return self._keys[:]
 
+    def __len__(self):
```
shoyer: Consider making `StructureType` inherit from `collections.Mapping` as well. That guarantees you don't forget anything, and helps for type checks (e.g., with `isinstance` or `typing.Mapping`).

src/pydap/model.py
```diff
@@ -568,6 +571,20 @@ def __getitem__(self, key):
...
+
+    @property
+    def dtype(self):
+        """Return the first children dtype."""
+        return self.array.dtype
+
+    @property
+    def shape(self):
+        """Return the first children shape."""
+        return self.array.shape
```
shoyer: It's not strictly necessary, but I usually make my duck array types implement `__array__`, `ndim` and `size` as well:
https://github.com/pydata/xarray/blob/371d034372bc7522098a142a0debf93916c49102/xarray/core/utils.py#L406

(Implement these on `BaseType`, too, if you decide to go this route)

src/pydap/model.py
laliberte: done

src/pydap/model.py
laliberte: And done!

src/pydap/model.py
laliberte: I've made it into a ``MutableMapping`` because this is what it is. Sounds good to you?

laliberte:
@shoyer could you confirm that ``MutableMapping`` is as good (if not better than) as ``Mapping``? If this is all good, anyone has anything else to say before I merge?

shoyer:
> @shoyer could you confirm that MutableMapping is as good (if not better than) as Mapping? If this is all good, anyone has anything else to say before I merge?

Be a little careful with MutableMapping -- it adds quite a few methods like pop, popitem, clear, update, and setdefault. It's probably fine in this case, but it's not desirable for every mutable Mapping type (e.g., xarray.Dataset is a collections.Mapping instance, but not a collections.MutableMapping).

laliberte:
Ok, that's a good point. Let's be on the safe side and use Mapping.

PR #49, opened 2017-01-03
Fix mechanicalsoup browser clean-up with error

laliberte:
This is a rebased version of #40.


laliberte:
I've have just rebased and this one should also be ready for merging.

PR #55, opened 2017-01-20
Fix tests and prevent unintended behaviour in numpy DeprecationWarning

### LINKED ISSUE #45: Namespace testcase fails to pass
In a fresh checkout of `pydap` in a clean venv, the `namespace` testcase does not pass:
```
======================================================================
FAIL: Test the namespace.
----------------------------------------------------------------------
  File "/home/gvaillan/.virtualenvs/pydap/src/pydap/src/pydap/tests/test_pydap.py", line 20, in test_namespace
    pydap.__doc__, "Declare the namespace ``pydap`` here.")
AssertionError: None != 'Declare the namespace ``pydap`` here.'

----------------------------------------------------------------------
Ran 311 tests in 0.582s
```
Indeed, if I manually run:
```python
>>> import pydap
>>> print(pydap.__doc__)
None
```
I am wondering whether this test case is actually useful, beyond increasing the code coverage?

laliberte:
This PR makes two minor fixes that are necessary to ensure tests are working:

1. Makes a copy of ``data`` when a structure is sliced with multiple names. This prevents the FutureDeprecationWarning from affecting ``pydap``.

2. Removes test for ``__doc__``. After extensive testing, it seems that python 3.5 always returns ``None``. For this reason and because it prevents a successful build I think it should be removed. It would also fix #45

### DIFF
--- a/src/pydap/__init__.py
+++ b/src/pydap/__init__.py
@@ -1,2 +1,5 @@
-"""Declare the namespace ``pydap`` here."""
+'''
+Declare the namespace ``pydap`` here.
+'''
+
 __import__('pkg_resources').declare_namespace(__name__)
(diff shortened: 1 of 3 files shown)


jameshiebert:
Sorry that it took so long to look this over.

PR #56, opened 2017-01-23
Convert all imports to explicit relative imports

laliberte:
This PR builds upon #55 to ensure that tests are successful.

This PR converts all imports to explicit relative imports. This makes it possible to include a fork of ``pydap`` as a submodule in another project. It is also much more robust as it prevents potential namespace collisions. Apart for the internal machinery, this PR should not change anything to the current ``pydap`` compatibility and capabilities.

### DIFF
--- a/src/pydap/__init__.py
+++ b/src/pydap/__init__.py
@@ -1,2 +1,5 @@
-"""Declare the namespace ``pydap`` here."""
+'''
+Declare the namespace ``pydap`` here.
+'''
+
 __import__('pkg_resources').declare_namespace(__name__)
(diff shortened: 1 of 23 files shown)


laliberte:
@jameshiebert Any chance this could be merged? There are relatively few risks involved it provides a much cleaner code base.

jameshiebert:
Looks pretty straightforward. Thanks for taking the time to do some code clean up and hope that the time investment pays off :)

laliberte:
Well, I kind of needed it for another project, so it already paid of!

PR #57, opened 2017-01-27
Fix to URS authentication

laliberte:
This PR fixes recent problems with the URS authentication.

The new implementation is based more closely on the ``requests`` library and should be much more robust in the future.

### DIFF
--- a/src/pydap/cas/esgf.py
+++ b/src/pydap/cas/esgf.py
@@ -17,7 +17,7 @@ def setup_session(openid, password, username=None,
                                         check_url=check_url,
                                         session=session,
                                         verify=verify)
-    # Connections can be lept alive on the ESGF:
+    # Connections can be kept alive on the ESGF:
     session.headers.update([('Connection', 'keep-alive')])
     return session
 
--- a/src/pydap/cas/esgf.py
+++ b/src/pydap/cas/esgf.py
@@ -11,12 +11,15 @@ def setup_session(openid, password, username=None,
 
     username should only be necessary for a CEDA openid.
     """
-    return get_cookies.setup_session(_uri(openid),
-                                     username=username,
-                                     password=password,
-                                     check_url=check_url,
-                                     session=session,
-                                     verify=verify)
+    session = get_cookies.setup_session(_uri(openid),
+                                        username=username,
+                                        password=password,
+                                        check_url=check_url,
+                                        session=session,
+                                        verify=verify)
+    # Connections can be lept alive on the ESGF:
+    session.headers.update([('Connection', 'keep-alive')])
+    return session
 
 
 def _uri(openid):
--- a/src/pydap/cas/esgf.py
+++ b/src/pydap/cas/esgf.py
@@ -4,7 +4,7 @@ from . import get_cookies
 
 def setup_session(openid, password, username=None,
                   check_url=None,
-                  session=None):
+                  session=None, verify=False):
     """
     A special call to get_cookies.setup_session that is tailored for
     ESGF credentials.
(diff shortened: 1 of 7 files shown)


laliberte:
This PR fixes in addition the ESGF auth tests. These tests were failing because the testing resource had disappeared.

src/pydap/cas/esgf.py
```diff
@@ -4,19 +4,22 @@
...
-                                     check_url=check_url,
-                                     session=session,
-                                     verify=False)
+    session = get_cookies.setup_session(_uri(openid),
+                                        username=username,
+                                        password=password,
+                                        check_url=check_url,
+                                        session=session,
+                                        verify=verify)
+    # Connections can be lept alive on the ESGF:
```
jameshiebert: lept=kept

src/pydap/cas/get_cookies.py
```diff
@@ -24,7 +24,10 @@ def setup_session(uri,
     '''
 
     if session is None:
-        headers = [('User-agent', pydap.lib.__version__)]
+        # Connections must be closed since some CAS
+        # will cough when connections are kept alive:
+        headers = [('User-agent', pydap__version__),
```
jameshiebert: Should this be ``'Pydap/{}'.format(pydap__version__)``, e.g. "Pydap/3.2.0" to conform to the [user-agent RFC](http://www.ietf.org/rfc/rfc2616.txt) (see section 3.8).

src/pydap/net.py
```diff
@@ -29,44 +31,39 @@ def raise_for_status(response):
...
-                    cookies_dict=None):
+def follow_redirect(url, application=None, session=None):
     """
     This function essentially performs the following command:
     >>> Request.blank(url).get_response(application)
 
     It however makes sure that the request possesses the same cookies and
     headers as the passed session.
-
-    It recursively follows 302 redirects.
```
jameshiebert: What, may I ask, is the key change here? You've removed the 302 following, in exchange for a HEAD request and copying the headers? If that does the job, great, but what headers does NASA/URS set that makes this possible? Would you be able to document that in the code, because it's essentially an external precondition that we'll require to be true. If that ever ceases to be true (and you're not involved in the development), that will be important for other people to know to be able to identify what has gone wrong.

laliberte: In fact this change was made to make future development easier. Here's the thinking:
1. ``get_cookies`` set the proper cookies in ``session``. This is checked using the ``check_url`` option, so we know it works.
2. ``requests`` is so well maintained that there are no reason why querying the url should fail: after all, we have checked that the credentials work.
3. ``requests`` handles all the cookies and the redirects. We simply have to use its result and pass it to ``webob`` for the last mile.

Makes sense? I can document that in the code if it does.

src/pydap/net.py
```diff
@@ -46,6 +46,16 @@ def follow_redirect(url, application=None, session=None):
...
+        # If session is set and cookies were loaded using pydap.cas.get_cookies
+        # using the check_url option, then we can legitimately expect that
+        # the connection will go through seamlessly. However, there might be
+        # redirects that might want to modify the cookies. Webob is not
+        # really up to the task here. The approach used here is to
+        # piggy back on the requests library and use it to fetch the
+        # head of the requested url. Requests will follow redirects and
+        # adjust the cookies as needed. We can then use the final url and
+        # the final cookies to set up a webob Request object that will
+        # be guaranteed to have all the needed credentials:
```
jameshiebert: This helps quite a bit. I think most of what I was looking for is what exactly does URS *do* that we need to work so hard to work around, but this at least explains how this code handles the redirects. Thanks.

laliberte: My idea was not to focus on URS but to provide a general method that would always work. Combined with the URS tests, this should be sufficient...

PR #59, opened 2017-01-27
Do not require gunicorn for a vanilla install

ocefpaf:
@rsignell-usgs reported issues on Widnows when trying to use `pydap`.
See https://gist.github.com/rsignell-usgs/7c9cf9f88a2f0c920c1a4ee7d7eaf601

This PR crates a `server_extras` entry to allow `pydap` to be installed, and usable without, `gunicorn` present.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,16 @@
 language: python
+
+sudo: false
+
 python:
     - 2.7
     - 3.3
     - 3.4
     - 3.5
-addons:
-  apt:
-    packages:
-      - gfortran
-      - libblas-dev
-      - liblapack-dev
-      - cython
+    - 3.6
+
 install:
-    - travis_wait 30 pip install scipy
+    - pip install scipy
     - pip install -e .[testing,functions,tests]
 
 script:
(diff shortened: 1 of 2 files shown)


jameshiebert:
This looks like a good idea to me (although we do seem to be acquiring a large collection of "extras"). One brief comment, if you set the Python 3.6 classifier (good idea!), can you ensure that py36 is also being testing on TravisCI?

ocefpaf:
> This looks like a good idea to me (although we do seem to be acquiring a large collection of "extras").

Indeed. Would you like to put the `gunicorn` dependency into one of the existing `extras`?

> One brief comment, if you set the Python 3.6 classifier (good idea!), can you ensure that py36 is also being testing on TravisCI?

Done.

jameshiebert:
> Would you like to put the gunicorn dependency into one of the existing extras?

No, I think a "server_extras" makes sense.

Hang tight on that last failed test. I'm going to merge one PR that should take care of that and then you can rebase.

jameshiebert:
OK, if you rebase this PR onto master, the full suite of tests should pass on Travis.

ocefpaf:
> OK, if you rebase this PR onto master, the full suite of tests should pass on Travis.

Done.

jameshiebert:
Perfect, thanks!

PR #60, opened 2017-01-27
Default charset to 'utf8'

ocefpaf:
URLs like the one below,

```python

url = 'http://geoport-dev.whoi.edu/thredds/dodsC/estofs/atlantic'
ds = open_url(url)
```

do not have a charset and will fail. It seems that it is safe to set a default to `utf-8`.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -49,11 +49,15 @@ class DAPHandler(BaseHandler):
         ddsurl = urlunsplit((scheme, netloc, path + '.dds', query, fragment))
         r = GET(ddsurl, application, session)
         raise_for_status(r)
+        if not r.charset:
+            r.charset = 'utf8'
         dds = r.text
 
         dasurl = urlunsplit((scheme, netloc, path + '.das', query, fragment))
         r = GET(dasurl, application, session)
         raise_for_status(r)
+        if not r.charset:
+            r.charset = 'utf8'
         das = r.text
 
         # build the dataset from the DDS and add attributes from the DAS
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -50,14 +50,14 @@ class DAPHandler(BaseHandler):
         r = GET(ddsurl, application, session)
         raise_for_status(r)
         if not r.charset:
-            r.charset = 'utf8'
+            r.charset = 'ascii'
         dds = r.text
 
         dasurl = urlunsplit((scheme, netloc, path + '.das', query, fragment))
         r = GET(dasurl, application, session)
         raise_for_status(r)
         if not r.charset:
-            r.charset = 'utf8'
+            r.charset = 'ascii'
         das = r.text
 
         # build the dataset from the DDS and add attributes from the DAS


ocefpaf:
@jameshiebert we have been using this patch with success in our package. See https://github.com/conda-forge/pydap-feedstock/pull/5/files#diff-e178b687b10a71a3348107ae3154e44cR13

Would you be OK merging this?

jameshiebert:
The only thing giving me pause is that (to my knowledge) [the DAP 2.0 spec](https://cdn.earthdata.nasa.gov/conduit/upload/512/ESE-RFC-004v1.1.pdf) is only actually defined for an ASCII encoding. Which I realize is kind of from last decade, but...
It *probably* doesn't matter in this context, and I'm happy to be further convinced, I'm just having trouble thinking through the implications (since the ``pydap.handlers.dap`` interfaces with both the client and the server). Has this been tested on both the client and the server? Or any I thinking too hard about this? :)

ocefpaf:
> The only thing giving me pause is that (to my knowledge) the DAP 2.0 spec is only actually defined for an ASCII encoding. 

I am not an expert on this but I cannot access that endpoint with our conda package without this patch.
Here are the steps to reproduce it:

```shell
conda create -n TEST pydap=3.2.0=py35_0
source activate TEST
```

and then inside the interpreter:

```python

url = 'http://geoport-dev.whoi.edu/thredds/dodsC/estofs/atlantic'
ds = open_url(url)

  File "<stdin>", line 1, in <module>
  File "/home/filipe/miniconda3/envs/TEST/lib/python3.5/site-packages/pydap/client.py", line 59, in open_url
    dataset = DAPHandler(url, application, session).dataset
  File "/home/filipe/miniconda3/envs/TEST/lib/python3.5/site-packages/pydap/handlers/dap.py", line 52, in __init__
    dds = r.text
  File "/home/filipe/miniconda3/envs/TEST/lib/python3.5/site-packages/webob/response.py", line 420, in _text__get
    "You cannot access Response.text unless charset is set")
AttributeError: You cannot access Response.text unless charset is set
```

I cannot reproduce that with the `PyPI` package, so maybe the issue lies with some of our dependencies? Why it happens with only a few URLs puzzles me as well.

> Or any I thinking too hard about this? :)

I understand your thinking, but this patch would fall back to 'utf-8' only when it fails to access the charset.
I chose `utf-8` b/c that was the recommendation I found when googling the problem. Maybe I should update the PR and use `ascii` instead?

ocefpaf:
I modified this PR to and our patch from https://github.com/conda-forge/pydap-feedstock/pull/6 to use `ascii` instead of `utf8`. Everything worked. (But I am under the impression that anything that is no `None` would do the trick here :smile:)

jameshiebert:
OK, sounds good. I agree that generally `utf-8` would be the safest default, but in the context of OPeNDAP, that's not the case (to my knowledge).

As to why this only happens with specific URLs, it's likely due to the fact that most HTTP servers will set the Content-Type: charset header, while some may not. If you look at the HTTP response from `geoport-dev.whoi.edu`, it is one of the unlucky few who do not:

```
$ curl -I http://geoport-dev.whoi.edu/thredds/dodsC/estofs/atlantic
HTTP/1.1 400 Bad Request
Date: Wed, 22 Feb 2017 18:29:09 GMT
Server: Apache-Coyote/1.1
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, OPTIONS
Access-Control-Allow-Headers: X-Requested-With, Content-Type, Origin, Authorization, Accept, Client-Security-Token, Accept-Encoding
Content-Description: dods-error
Content-Type: text/plain
Content-Length: 65
Vary: Accept-Encoding
Connection: close
```

And the `WebOb` client library needs to know what the charset is for a programmer to be able to use the `text` property: http://docs.webob.org/en/stable/api/response.html#webob.response.Response.text

ocefpaf:
Awesome! Thanks!!

Can get a new release with the fixes :grimacing: ?

PR #61, opened 2017-02-05
Treat '-nan' attribute value as a nan value

MBARIMike:
I encountered some EPIC-formatted data that contains '-nan' values in some attributes. This PR fixes a problem where pydap would raise a *** ValueError: malformed string Exception.


jameshiebert:
That seems reasonable to me. Thanks!

MBARIMike:
Thank you!

PR #68, opened 2017-02-26
Add the test data to the release tarball

### LINKED ISSUE #54: Tests cannot be run standalone
Despite the fact that `pydap.tests` is provided in the PyPI tarball, the test suite cannot be run for the following reasons:

- Missing data: *.das, *.dds, *.dods. These should be added to the `package_data` parameter of the setup object in `setup.py`.

- Reliance on `Pydap.egg-info`. This prevents the possibility of running the tests in a standalone way by doing:
```
python setup.py build
cd build/lib
python -m nose -a '!auth'
```

This is particularly desirable for packaging purposes (Debian for instance), where the package is built for all supported Python versions concurrently. Each Python version has its own build directory in which the test suite is run. Right now, this is not possible for these 2 reasons.

ghisvail:
The test suite provided within the release tarball cannot be run successfully without the data. Closes #54

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,7 @@
 recursive-include docs *
 recursive-include src/pydap/wsgi/templates *
 recursive-include src/pydap/responses/html/templates *
+recursive-include src/pydap/tests *.das *.dds *.dods
 include CONTRIBUTORS.md
 include README.md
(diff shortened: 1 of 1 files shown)


jameshiebert:
Awesome, thanks for fixing this.

Issue #75, opened 2017-03-07
ValueError: dictionary update sequence element #0 has length 1; 2 is required


mortenalver:
Hi,

I'm having problems with Pydap 3.2.0 installed on a 64 bit Linux system (however, have also seen the problem on Windows 10). The URL I try to open works fine under Pydap 3.1.1, but under 3.2.0 I get an error message. I can reproduce the error as shown below:
```
>>> from pydap.client import open_url
>>> dataset = open_url('http://thredds.met.no/thredds/dodsC/fou-hi/norkyst800m-1h/NorKyst-800m_ZDEPTHS_his.fc.2017030700.nc')          Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/malv/.local/lib/python3.5/site-packages/pydap/client.py", line 59, in open_url
    dataset = DAPHandler(url, application, session).dataset
  File "/home/malv/.local/lib/python3.5/site-packages/pydap/handlers/dap.py", line 61, in __init__
    add_attributes(self.dataset, parse_das(das))
  File "/home/malv/.local/lib/python3.5/site-packages/pydap/parsers/das.py", line 127, in add_attributes
    var.attributes.update(nested.pop(k))
ValueError: dictionary update sequence element #0 has length 1; 2 is required
>>>
```

phrrngtn:
I got a 404 error when I tried to run this just now.

mortenalver:
Hi,

Probably due to daily file name changes on the server. To get a working
link you can update the date in the link to today's. Not able to test at
the moment.

Morten


8. mar. 2017 7.10 p.m. skrev "Paul Harrington" <notifications@github.com>:

mortenalver:
This URL should work for the foreseeable future (the particular link I used was badly chosen, but this one should remain): 

http://thredds.met.no/thredds/dodsC/fou-hi/norkyst800m-1h/NorKyst-800m_ZDEPTHS_his.an.2017030800.nc

phrrngtn:
Thanks, that URL worked for me.  I took a quick look in the debugger and it looks like the time key is absent from zeta but I don't know if this is a problem with the response or the code.

ipdb> p nested.pop(k)
*** KeyError: KeyError('time',)
ipdb> var.id
'zeta.time'

laliberte:
@mortenalver if you want you can fork PR #75 and check whether it works for your dataset. On my side, it worked flawlessly.

mortenalver:
[...]

PR #79, opened 2017-03-24
Fix attributes parser error raised in #75

laliberte:
This PR solves #75. It introduces a more stringent DAS test.


laliberte:
You're right, it could be tightened. KeyError is on nested.pop(k) and the new exception is on var.attributes.update(value). I'm fixing this right now.

jameshiebert:
Cool, that should work. Thanks!

PR #80, opened 2017-03-29
Remove scipy from testing requirements

ghisvail:
`scipy` is only needed transitively by `gsw` (upstream made this dependency [optional](https://github.com/TEOS-10/python-gsw/commit/2fb1ab7518e2797e690c895a8d501e60f368b64d) since the last release).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,6 @@ cas_extras = [
 tests_require = functions_extras + cas_extras + server_extras + [
     'WebTest',
     'beautifulsoup4',
-    'scipy',
     'flake8'
 ]
 


jameshiebert:
OK cool. That should speed up our build. Can you pull it out of the [TravisCI config](https://github.com/pydap/pydap/blob/master/.travis.yml#L19) for this PR as well?

laliberte:
I think you could remove all ``addons`` AND the ``pip install scipy`` line. This will be much a needed speedup for Travis-CI!

ghisvail:
Please go ahead a merge the PR and feel free to follow it by a commit to address the CI speed-up.

laliberte:
Ok, will do!

Issue #76, opened 2017-03-10
How can I read array of Structures?


hyoklee:
It seems that PyDAP doesn't provide a way to access data from array of structures.
``
Dataset {
    Structure {
        Float64 Time;
        Float32 Latitude;
        Float32 Longitude;
      }  Point[13356];
}
``

Here's a real test case:

https://eosdap.hdfgroup.org:8989/opendap/hyrax/data/NASAFILES/hdf5/AMSR_U2_L2_Land_B01_201207232336_D.he5.html

Shape for structure returns an error. I tried open_dods() but it did not work, either. Is there a working solution for accessing array of structure data?

laliberte:
Unfortunately, after some tests, I must conclude that this type of dataset is not currently supported by ``pydap``. It will require substantial code modification to work so I think it is unlikely to become available in the near future. I'd say that this issue should remain opened until a suitable code modification is made available.

laliberte:
I have been having a look at this issue over the last few days and made somewhat little headway. This is made difficult by the fact that the current data model always grabs data from the simplest datatype in the hierarchy. However, this example server show clearly that this is not how the DAP2 standard should work: one cannot fetch subcomponents of a ``StructureType`` that has a ``shape``; instead one has to fetch all subcomponents of said ``StructureType`` for a slice of the data and THEN extract the desired subcomponent.
This will require a fairly comprehensive re-organisation of the ``pydap`` data model. If my preliminary tests are right, it might be possible to do this without changing much of the public API (at least what seems to be implicitly considered a public API). I expect this should take a couple of weeks, if things go smoothly...

hyoklee:
Thanks @laliberte for working hard on this issue.

PR #81, opened 2017-03-30
Parse dods response to handle ERDDAP servers

### LINKED ISSUE #6: Test Erddap responses
Some Erddap servers return invalid DDS responses when a deep object is returned. In that case, we should parse the DDS from the DODS response and extract the data using it.

laliberte:
This PR fixes #6. This code should now work:
```

dataset = open_url('http://oos.soest.hawaii.edu/erddap/griddap/NCEP_Global_Best')
tmpsfc = dataset['tmpsfc']
tmpsfc[0, :, :]
```

This code also makes it possible to solve #76. The way I see it, it will only require a few tweaks to allow these types of links.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ tests_require = functions_extras + cas_extras + server_extras + [
 
 testing_extras = tests_require + [
     'nose',
+    'mock',
     'coverage',
     'requests'
 ]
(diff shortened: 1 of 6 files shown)


laliberte:
@jameshiebert do you give your thumbs up for this to be merged?

jameshiebert:
Yeah, I like the way that you split out the utility functions. Makes the code more clear. In any case, it looks good from my perspective.

laliberte:
Ok, I'll go ahead and merge!

PR #83, opened 2017-03-30
Basic pytest compatiblity

laliberte:
This PR converts testing to ``pytest``. It does not modify other aspects of testing and further work will have to be done to modernize tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,5 +20,5 @@ install:
     - pip install -e .[testing,functions,server,tests]
 
 script:
-    - python setup.py nosetests -a '!auth'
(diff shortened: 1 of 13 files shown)


laliberte:
@jameshiebert if you give your thumbs up, I will merge. This will close #83 but I hope this won't be the end of the discussion.

jameshiebert:
Wow, is that all there is to it? I would have done this a long time ago :)

laliberte:
I had the same impression but I tried this morning and it worked. I guess the ``pytest`` team has worked extra hard to enable such a transition! Good for us :)

PR #85, opened 2017-03-30
Give default name to model components to fix #42

### LINKED ISSUE #42: model.BaseType required name argument
I just upgraded from Pydap 3.1.1 to Pydap 3.2 from PyPI and my unit tests failed with:

``` 
 File "/home/travis/build/MBARIMike/stoqs/stoqs/loaders/DAPloaders.py", line 1362, in Dorado_Loader
    chl = pydap.model.BaseType()
TypeError: __init__() takes at least 2 arguments (1 given)
```

Version 3.1.1 provides a default value for name:

```python
    def __init__(self, name='nameless', data=None, shape=None,
            dimensions=None, type=Int32, attributes=None):
        DapType.__init__(self, name, attributes)
```

Is there any harm in providing this same default value version 3.2 ([this line](https://github.com/pydap/pydap/blob/a7ca9c58f4d628fb1fc232a3c5653e6f2ab9378e/src/pydap/model.py#L199)) so that we maintain backward compatibility?

laliberte:
This PR fixes #42. It does not break any tests so I don't see why we would not provide this backward-compatibility.

What do you think @jameshiebert ?

### DIFF
--- a/src/pydap/model.py
+++ b/src/pydap/model.py
@@ -191,7 +191,7 @@ class DapType(object):
 
     """
 
-    def __init__(self, name, attributes=None, **kwargs):
+    def __init__(self, name='nameless', attributes=None, **kwargs):
         self.name = quote(name)
         self.attributes = attributes or {}
         self.attributes.update(kwargs)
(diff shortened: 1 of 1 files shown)


laliberte:
@jameshiebert any opinion on this one?

laliberte:
If no one else has comments, I'm going to merge tomorrow.

PR #86, opened 2017-03-30
Fix a bug where the iteration does not replace previous_chunk

laliberte:
This is a weird bug that I only picked up when working on a different PR. Seems straighforward.


laliberte:
@jameshiebert this one seems straighforward. I think it should be merged. Any opinion?

laliberte:
Again, If no one else has comments, I'm going to merge tomorrow.

PR #87, opened 2017-03-30
Drop usage of Webob in favor of requests

laliberte:
I know the opposite was done in commit 91f01609ec3536279cf985695cfab6eca2124374 but ``requests`` is so much better than ``webob`` to handle redirects and authentication. In this PR, I propose converting ``requests`` to a WSGI using an adapter. I seems to be working. This is not urgent and the goal here is to have a discussion.

What's your gut feeling @jameshiebert ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,8 @@ install_requires = [
     'docopt',
     'six >= 1.4.0',
     'mechanicalsoup',
+    'requests',
+    'requests_wsgi_adapter'
 ]
 
 if sys.version_info < (3, 5):
@@ -31,10 +33,6 @@ docs_extras = [
     'sphinx_rtd_theme',
 ]
 
-cas_extras = [
-    'requests'
-    ]
-
 tests_require = functions_extras + cas_extras + server_extras + [
     'WebTest',
     'beautifulsoup4',
@@ -45,8 +43,7 @@ tests_require = functions_extras + cas_extras + server_extras + [
 testing_extras = tests_require + [
     'pytest-cov',
     'pytest-attrib',
-    'mock',
-    'requests'
+    'mock'
 ]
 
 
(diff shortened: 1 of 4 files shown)


jameshiebert:
So, my main gut feeling is to be slightly wary.

I *do* like the ``requests`` library. It is a beautiful piece of software that is easy to use and provides tons of great features for the client side. Given all of your work on the redirects and authentication, I can definitely understand why it would be desirable to utilize it.

But Pydap includes code for both the client and the server. It's written at the WSGI level, so to me it makes sense to use a helper library that's a bit lower level and has direct hooks into the WSGI interface.

``WebOb`` is very mature and is well used and supported by the Pylons team, so I trust its code quality. Whereas the ``requests-wsgi-adapter`` packages is technically [still beta](https://github.com/seanbrant/requests-wsgi-adapter/blob/master/setup.py#L26). I mean, there's not a lot of code there, so not a lot to go wrong. But on the other hand, I kind of wonder whether 150 LOCs can handle all of the various ways in which Pydap uses WSGI on both the client and server.

It also seems like a kind of backward abstraction. We would use ``requests`` to interact with our Pydap code and then use the ``requests-wsgi-adapter`` to get access to the WSGI objects? It feels weird, like we're adding two layers of code to be able to work a lower level (then again, maybe I misunderstand how the adapter works?).

So, personally, I'm kind of skeptical, but of course I see some value in having higher level code, so if we can do it in a consistent manner, I'm happy to see a proof-of-concept. I see about two dozen webob imports across the code and tests, so it may take a bit of work to ensure that ``requests``+``requests-wsgi-adapter`` can actually replace all of them.

Does this make any sense?

laliberte:
Thanks for your insight!

I understand your argument and I must admit that you convinced here. I find ``webob`` cumbersome and not always very user friendly but it seems that, at least for the foreseeable future, it would be much better to stick with it. In a ways, it is much more elegant and easier to maintain to use ``webob`` as the main engine for both requests and responses. Any other option would require to have to distinct engines and this could easily become a nightmare to maintain.

In short, I'll close this PR but maybe in a couple of years it could be revisited, who know?

Issue #3, opened 2013-10-24
Move `_original_keys` to `pydap.model.SequenceType`.


robertodealmeida:
Right now `IterData` tracks the original key order by setting an attribute called `_original_keys` in the sequence. This behavior should be moved to the original class, to keep the code more simple.

laliberte:
I'm going to close this.

PR #89, opened 2017-04-03
Fix Mapping scheme for SequenceType

laliberte:
This PR is a follow-up on #48. The goal here is to ensure that all DataTypes defined in the ``pydap`` model follow the ``Mapping`` semenatics where iterations are over children and not over data. This feature was broken by the implementation of an iteration over data in the ``SequenceType``. I strongly argue for this feature to be dropped. This would be in line with other pydata projects like ``xarray``. Here, I propose that the iteration over data be replaced by an explicit call to a new ``.iterdata()`` method. There could other ways of transitioning to a more explicit iteration over ``SequenceType`` data but it felt like the simplest. @jameshiebert, @shoyer does any of you have a suggestion?

One positive consequence of this PR is that ``pydap`` links can now be decoded through ``xarray`` (basic tests work with this PR). This would solve this [issue](https://github.com/pydata/xarray/issues/1174) and would likely allow progress on this other [issue](https://github.com/pydata/xarray/issues/1068)


src/pydap/tests/test_model.py
```diff
@@ -299,76 +301,83 @@ def test_id(self):
         self.assertEqual(child.id, "child")
 
 
-class TestSequenceType(unittest.TestCase):
+@pytest.fixture
```
laliberte: I have started switching the tests to ``pytest`` when useful. In this case, I was only able to get the ``DeprecationWarning`` test to succeed with this approach.

src/pydap/model.py
```diff
@@ -307,15 +305,45 @@ def __getattr__(self, attr):
...
             if isinstance(x, binary_type):
-                yield x.tostring().decode('utf-8')
+                x = x.tostring().decode('utf-8')
+            return x
+        except KeyError:
+            splitted = key.split('.')
+            if len(splitted) > 1:
+                try:
+                    return (self
+                            .__getitem__(splitted[0])['.'.join(splitted[1:])])
```
shoyer: use `self[...]` instead of `self.__getitem__(...)`

src/pydap/model.py
```diff
@@ -526,18 +530,18 @@ def __getitem__(self, key):
 
         # If it's a tuple, return a new `SequenceType` with selected children.
         elif isinstance(key, tuple):
-            out = SequenceType(self.name, self.data, self.attributes.copy())
+            out = self.__class__(self.name, self._data, self.attributes.copy())
```
shoyer: nit: usually I write `type(self)` instead of `self.__class__`, but I think they do the same thing.

src/pydap/model.py
```diff
@@ -287,18 +288,15 @@ class StructureType(DapType, Mapping):
     """A dict-like object holding other variables."""
 
     def __init__(self, name, attributes=None, **kwargs):
-        DapType.__init__(self, name, attributes, **kwargs)
+        super(StructureType, self).__init__(name, attributes, **kwargs)
 
         # emulate a simple ordered dict
```
shoyer: just a note: if you're not supporting Python 2.6 (which you shouldn't be, it's ancient!), then consider switching this to actually just use an OrderedDict.

src/pydap/model.py
```diff
@@ -307,15 +305,45 @@ def __getattr__(self, attr):
...
+    # and __ne__ are obtained.
     def __iter__(self):
+        if isinstance(self, SequenceType):
+            warnings.warn('Iteration now yields children names. '
+                          'This means that ``for val in sequence: ...`` '
+                          'will give children names. '
+                          'To iterate over data use the construct '
+                          '``for val in sequence.iterdata(): ...``',
+                          DeprecationWarning)
         for key in self._keys:
```
shoyer: just use `return iter(self._keys)`

src/pydap/model.py
```diff
@@ -287,18 +288,15 @@ class StructureType(DapType, Mapping):
...
 
         # emulate a simple ordered dict
         self._keys = []
         self._dict = {}
 
     def __repr__(self):
         return '<%s with children %s>' % (
-            self.__class__.__name__, ', '.join(map(repr, self.keys())))
-
-    def __contains__(self, child):
```
shoyer: It's still a good idea to implement `__contains__`. The `collections.Mapping` implementation defaults to iterating over every element.

src/pydap/model.py
```diff
@@ -290,13 +290,13 @@ class StructureType(DapType, Mapping):
     def __init__(self, name, attributes=None, **kwargs):
         super(StructureType, self).__init__(name, attributes, **kwargs)
 
-        # emulate a simple ordered dict
-        self._keys = []
+        # allow some keys to be hidden:
+        self._visible_keys = []
```
shoyer: If you don't need to keep track of order, use a set instead of a list.

laliberte: Good point, I'll change it. How do you like the new approach to dealing with keys?

shoyer:
It would be helpful if you could include two summaries:

1. How the types you are changing worked before (especially in pydap 3.1)
2. How the types work how.

src/pydap/model.py
shoyer: I'm still sorting through what this change means, exactly.

laliberte:
The latest commit uses an ``OrderedSet`` and fixes #3 coincidentally. ``test_model.py`` has now been converted to ``pytest`` semantics for consistency (Deprecation testing did not work with ``unittest``). Finally, the latest commit does not deprecate the iteration as before. Instead, it provides a compromise where iteration is over data but ``items``, ``values`` and ``keys`` are over children and therefore likely to be more forward-compatible. A ``PendingDeprecationWarning`` is now issue to allow us to switch to a pure ``Mapping`` in the future.

shoyer:
I'll need to read up on the pydap data model to be able to give a useful review here. I don't really understand it right now.

laliberte:
@shoyer I'll try to document how the new datatypes work but I'm really not an expert on the 3.1 model...

shoyer:
The developer docs seem to explain things pretty thoroughly:
http://www.pydap.org/en/latest/developer.html#data

On Tue, Apr 4, 2017 at 1:44 PM, Frederic Laliberte <notifications@github.com

laliberte:
Perfect, and that seems to be the ``pydap`` 3.1 datamodel. Do you think I should put the description in the docs or directly in the code?

shoyer:
The docs should definitely be updated as well as the code.

On Tue, Apr 4, 2017 at 1:47 PM, Frederic Laliberte <notifications@github.com

src/pydap/model.py
```diff
@@ -512,32 +532,76 @@ def _get_data(self):
...
+                             'This means that in the future '
+                             '``for val in sequence: ...`` '
+                             'will give children names. '
+                             'To iterate over data the construct '
+                             '``for val in sequence.iterdata(): ...``'
+                             'is available and will be supported in the'
+                             'future')
+
+    def __iter__(self):
+        # This method should be removed once the deprecation is
```
jameshiebert: For all of these future deprecation comments, I think that it should be specified in the comment what the target for deprecation is. If somehow these get overlooked during the course of another couple releases, it would be nice to have it say exactly what our intent was, so that we can compare that to the future present. :) E.g. if the comment says "remove this by version 3.4", and we're on version 3.5, it's obvious to even the casual observer that action is required.

laliberte: Sounds fair to me. Any suggestion?

laliberte:
FYI, @jameshiebert the new commit is backward-compatible. All it does is provide the ``iterdata()`` method and issue a ``PendingDeprecationWarning`` (that is difficult to test properly in 2.7 and 3.3, the reason why this is broken on Travis). It also ensures that ``SequenceType`` is mostly compatible with a ``Mapping`` by providing a custom-made compatibility layer (methods ``items``, ``keys``, ``values`` and ``__contains__``) that should be removed once iteration over data is deprecated.
That being said, it appears to be changing very little to the API so could probably be merged directly into master. I would however be completely comfortable with a minor release to ensure that users are made aware of these improvements.

jameshiebert:
Right, right. Sorry, I spaced out on that :)

laliberte:
@jameshiebert don't worry, it's a big PR. At least bigger than I originally expected!

laliberte:
The latest commit does:

1. Update the docstrings and the developer's documentation.
2. Both docstrings and the documentation are now tested using ``doctest`` through ``pytest``.
3. Some files were excluded in ``conftest.py`` because they use external links.
4. Everything builds fine on Travis.

@shoyer: I tried to make ``visible_keys`` a set but it became cumbersome in iterations to ensure order. Is there some obvious tricks I'm not aware of?

src/pydap/model.py
laliberte: This was done.

src/pydap/model.py
laliberte: done

src/pydap/model.py
laliberte: I've set ``pydap`` 3.4 as the deprecation version.

src/pydap/model.py
```diff
@@ -37,21 +55,54 @@
...
+Slicing these datasets with a list of keywords yields a `StructureType`
+or `DatasetType` with only a subset of the children::
+
+    >>> print(dataset.s['foo', 'foobar'])
+    <StructureType with children 'foo', 'foobar'>
+    >>> print(list(dataset.s['foo', 'foobar'].keys()))
+    ['foo', 'foobar']
+
+In the same way, the ``.items()`` and ``.values()`` methods are like in python
+dictionnaries and they iterate over sliced values.
```
shoyer: spelling: dictionnaries -> dictionaries

src/pydap/model.py
```diff
@@ -37,21 +55,54 @@
...
+
+The `StructureType`` never forgets of its children when sliced. Slicing a
+`StructureType` in fact simply changes its ``visible_keys`` property::
+
+    >>> print(dataset.s['foo', 'foobar'].visible_keys)
```
shoyer: This feels like a complex implementation detail that doesn't need to be exposed to users. What's the use for pulling out `visible_keys` rather than just looking at the dataset? Why use `all_keys()` or `all_values()` instead of keeping around the original object?

src/pydap/model.py
```diff
@@ -307,55 +392,100 @@ def __getattr__(self, attr):
...
-    def keys(self):
-        """Method to emulate a dictionary, returning keys."""
-        return self._keys[:]
+    def __getitem__(self, key):
+        if isinstance(key, string_types):
+            return self._getitem_string(key)
+        elif (isinstance(key, tuple) and
+              all(isinstance(name, string_types)
+                  for name in key)):
+            self.visible_keys = list(key)
```
shoyer: Mutating objects in `__getitem__` is very surprising. Users expect a new object, which I would construct similarly to how it's done in `_getitem_string_tuple` above by copying the listed keys.

src/pydap/model.py
```diff
@@ -287,18 +375,15 @@ class StructureType(DapType, Mapping):
...
 
     def __init__(self, name, attributes=None, **kwargs):
-        DapType.__init__(self, name, attributes, **kwargs)
+        super(StructureType, self).__init__(name, attributes, **kwargs)
 
-        # emulate a simple ordered dict
-        self._keys = []
-        self._dict = {}
+        # allow some keys to be hidden:
+        self._visible_keys = []
```
shoyer: Based on reading through what this class does, I don't think you actually need a `_visible_keys` attribute. Instead, you can simply remove or add items from `self._dict`, which is now ordered.

src/pydap/model.py
```diff
@@ -307,55 +392,100 @@ def __getattr__(self, attr):
...
+            out[name] = copy.copy(self._getitem_string(name))
+        return out
 
-    def keys(self):
-        """Method to emulate a dictionary, returning keys."""
-        return self._keys[:]
+    def __getitem__(self, key):
+        if isinstance(key, string_types):
+            return self._getitem_string(key)
+        elif (isinstance(key, tuple) and
```
shoyer: It might be a better idea to use `list` instead of `tuple` for this special type. Here's why: Python translates `x[y, z]` into `x.__getitem__((y, z))`, automatically turning multiple arguments in `[]` into a tuple. So it would be harder to use `[]` accidentally.

That said, let's save this one for later, given that apparently this was supported in earlier versions of pydap.

src/pydap/responses/dds.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, dataset):
 
     def __iter__(self):
         for line in dds(self.dataset):
+            print(line)
```
shoyer: should be removed

src/pydap/net.py
```diff
@@ -34,7 +34,7 @@ def raise_for_status(response):
 def follow_redirect(url, application=None, session=None):
     """
     This function essentially performs the following command:
-    >>> Request.blank(url).get_response(application)
+    >>> Request.blank(url).get_response(application) #doctest: +SKIP
```
shoyer: PEP8 guidelines for inline comments are to always preface them by at least two lines and include a space after `#`. So this should be:
```
>>> Request.blank(url).get_response(application)  # doctest: +SKIP
```

shoyer:
After reading over the old/new docs, I finally understand what's going on here!

I think the fundamental challenge here is that `SequenceType` is a funny type. It tries to be both array-like (like a NumPy array with a structured type) and dict-like (it inherits from `StructureType`, which is clearly only a Python mapping). It feels like the pydap version of a data frame (e.g., from pandas), in the way that it is both array-like and dict-like and designed for holding tabular data.

The challenge is that the array-like and dict-like properties come into conflict. I don't think it's possible to make an object that fully satisfies the expectation users would have about both types: dicts iterate over keys whereas arrays (including structured dtypes) iterate over rows.

That said, I do agree with the design decision here to switch the default behavior to be more dict-like, and add a separate method for iterating over rows. This makes more sense for the package as a whole, given that `SequenceType` inherits from `StructureType`. The more array-like option could also work, but in that case I would probably want to switch to inherit from `BaseType` instead.

docs/client.rst
```diff
@@ -12,131 +12,109 @@ Let's start accessing gridded data, i.e., data that is stored as a regular multi
...
 
 .. doctest::
 
-    >>> print sst.array[0,10:14,10:14] 
-    [[ -1.26285708e+00  -9.99999979e+33  -9.99999979e+33  -9.99999979e+33]
-     [ -7.69166648e-01  -7.79999971e-01  -6.75454497e-01  -5.95714271e-01]
-     [  1.28333330e-01  -5.00000156e-02  -6.36363626e-02  -1.41666666e-01]
-     [  6.38000011e-01   8.95384610e-01   7.21666634e-01   8.10000002e-01]]
-
-Instead of indexes we can also subset the data using its maps, in a more natural way. Just keep in mind that sometimes axes can be cyclic, like longitude, and you may have to download two separate parts and concatenate them together. This is not the case here:
```
laliberte: Unfortunately, it does not work anymore. It might be possible to reintroduce without too much pain but at the moment ``GridType`` does not support array slicing :(

src/pydap/model.py
laliberte: Yes, this was weird. Now, the object is not mutated but copied (with all of its children) and its ``_visible_keys`` attribute modified. See my other comment for why this has to be done (unfortunately).

src/pydap/model.py
laliberte: I agree with you that a list would be preferable and that it should be implemented later.

src/pydap/model.py
laliberte: That would be great but there are one with ``SequenceType``. One might want to slice based on field ``'a'`` but get ``['b', 'c']``:
``
data = sequence[sequence['a'] > 0.0]['b', 'c']
for s in data.iterdata():
    print(s)  # This is when data is retrieved and sliced.
``
Because a ``SequenceType`` is expecting streaming data, this slicing cannot be done now but later. This means that ``'a'`` must carried around until actual data is passed. This is burdensome but it seems that changing this at the moment would be a major undertaking.

src/pydap/model.py
laliberte: I agree that these methods could be hidden. I'll see how it looks like for my next commit.

docs/developer_data_model.rst
```diff
@@ -0,0 +1,415 @@
...
+
+    >>> a = BaseType('a', data=np.array([1]))
+    >>> print(a.data)
+    [1]
+
+    >>> b = BaseType('b', data=np.arange(4))
+    >>> print(b.data)
+    [0 1 2 3]
+
+Note that the default type for variables is ``Int32``:
```
laliberte: Updated to current behavior: infer from input data.

docs/developer_data_model.rst
```diff
@@ -0,0 +1,415 @@
...
+
+``BaseType``
+************
+
+For the simple ``BaseType`` objects the ``data`` attributes is usually a Numpy array, 
+though we can also use a Numpy scalar or Python number:
+
+.. doctest::
+
+    >>> a = BaseType('a', data=np.array([1]))
```
laliberte: done

src/pydap/model.py
shoyer: OK, fair enough. I didn't realize that `SequenceType` can be directly fed lazy data in the form of structured arrays (records) rather than separate 1D arrays for columns.

laliberte: I think this issue might have to be revisited. It is definitely not really clean. While it is cleaner now, I have the feeling that it might now belong in the ``StructureType`` class but in the ``SequenceType``. I want to be careful with this because it might come back haunt us if we want to implemented ``ConstraintExpressions`` (like the geographical constraints you inquired about) on ``StructureTypes``. At the moment, I would err on the safe side and leave it there but I want to have another look at the code, just to make sure.

laliberte:
You're right that ``SequenceType`` is essentially a ``DataFrame``. Ideally, over time, we would make ``SequenceType`` more and more similar to the ``pandas.DataFrame``. I however think we should leave this for another PR...

shoyer:
Agreed it's best to leave it for later. Honestly, I'm not convinced
SequenceType makes a whole lot of sense in pydap -- there are undoubtedly
much better options for accessing tabular data over a network than DAP. But
if you find it valuable, I encourage you to pursue it :).

On Fri, Apr 7, 2017 at 3:30 PM, Frederic Laliberte <notifications@github.com

laliberte:
That's what I think. Development in ``pydap`` should focus on ``GridType`` data while ``SequenceType`` should only be maintained (and tweaked as in here) to provide compatibility with the DAP2 standards.

src/pydap/model.py
```diff
@@ -368,6 +353,12 @@ def _get_data(self):
 
     def _set_data(self, data):
         self._data = data
+        if np.isscalar(data):
```
shoyer: Can you add a unit test that exercises this?

src/pydap/model.py
```diff
@@ -393,32 +384,25 @@ def __getattr__(self, attr):
...
     # From these, keys, items, values, get, __eq__,
     # and __ne__ are obtained.
     def __iter__(self):
         for key in self._dict.keys():
-            if key in self.visible_keys:
+            if key in self._visible_keys:
                 yield key
 
-    def all_keys(self):
+    def _all_keys(self):
```
shoyer: Are these methods necessary at all or can they simply be removed instead? It looks like you only had to update a direct test of this functionality.

laliberte: In fact ``_all_keys`` is the only required method. I've deleted ``_all_items`` and ``_all_values``.

src/pydap/model.py
laliberte: Yes and I've added full test coverage on this file.

laliberte:
Unless someone has anything else to add, I will merge this in a couple of hours.

PR #91, opened 2017-04-07
Unify dods parsers

laliberte:
When #81 was merged, we did not notice that there was two separate DODS parsers implementations. This means that some tests were not performed on both, limiting the scoping of our testing. This PR merges the two. The result is a single DODS parser and a cleaner code.


laliberte:
This PR is pretty straightforward and mostly simplifies the code. I have taken great care to ensure that the new unified parser has the same performances or better as before.
@jameshiebert do you have any comments? If not, I will be merging this PR early next week.

laliberte:
All checks are ok. I am merging this PR.

Issue #33, opened 2016-12-09
Launch a local pydap server from Python


shoyer:
For testing pydap read support in xarray, it would be nice to able to launch a pydap server locally as part of the test suite, avoiding the need to contact an external test server. For example:
```
expected_dataset = ...
with local_pydap_test_server() as port:
    actual_dataset = xarray.open_dataset('http://localhost:%s' % port, engine='pydap')
    assert actual_dataset.equals(expected_datset)
```

Is there a straightforward way to do this with the current version of pydap?

jameshiebert:
I think that most of our tests work one level down at the WSGI level and skip HTTP altogether. This is quite a bit faster and then avoids any problems with the network stack that you don't really want your tests to rely on..

We do this by passing a WSGI app object into WebOb.request.blank (IIRC). Would something like that work for you?

shoyer:
> We do this by passing a WSGI app object into WebOb.request.blank (IIRC). Would something like that work for you?

Sure, though I would prefer to be able to use a utility function from pydap rather than constructing such objects myself. WSGI and WebOb feel like implementation details that I would rather not leak into the xarray test suite.

laliberte:
@shoyer, @jameshiebert 
I have actually been working on this at the end of last week. It's in [merging/handlers.csv](github/laliberte/pydap/merging/handlers.csv).

To get this branch together, I wanted to clean-up the CSV handler so that I could have a real sever test. In that branch, the test server is in the file ``test_devel_server.py``.

I've also been working on a ``netcdf4`` [merging/handlers.netcdf](github/laliberte/pydap/merging/handlers.netcdf) handler to test more complex use case.

I was thinking that I could also code an ``xarray``-based handler. It would then make it possible to create a suite of unittests for ``xarray``-``pydap`` integration. Would that be something you would be interested in @shoyer ?

shoyer:
> I was thinking that I could also code an xarray-based handler. It would then make it possible to create a suite of unittests for xarray-pydap integration. Would that be something you would be interested in @shoyer ?

Indeed, an xarray based handler seems like a natural addition. I can see xarray (especially with dask) being a nice way to setup a pydap server, e.g., as a more flexible substitute for [pydap.handlers.nca](https://github.com/pydap/pydap.handlers.nca). This isn't something I would use personally (I would be more likely to just launch a remote dask-distributed server), but you should build it if you would find it useful.

[...]

PR #92, opened 2017-04-09
Simple devel test server

laliberte:
This PR provides the development test server from PR #34. The goal here is dissociate the merging of the CSV handler from the introduction of a test server.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,7 @@ functions_extras = [
 
 server_extras = [
     'gunicorn',
+    'werkzeug'
 ]
 
 docs_extras = [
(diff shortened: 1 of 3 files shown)


laliberte:
I feel this most recent attempt at tackling issue #33 is the cleanest approach possible and will provide a much needed feature for some related projects.

Where do you stand on this @jameshiebert ? Ideally, I would like to merge this PR and #94 by the end of next week.

laliberte:
@jameshiebert should I go ahead and merge this one?

jameshiebert:
Yeah, this seems pretty reasonable. The only thing that I'd say is that it's probably worth adding a section in the server docs (with a big caveat that says *this is not for production use*).

I'll merge this and create a separate issue for the docs.

PR #94, opened 2017-04-13
Add a timeout option to open_urls and open_dods

laliberte:
This PR follows #92 and supersedes #37. It introduces a simple timeout feature that enables a user to use ``open_urls`` and ``open_dods`` with an optional ``timeout`` argument. When opened this way, urls will raise an error is the server fails to respond is fewer seconds than specific by the ``timeout`` argument.

### DIFF
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -441,12 +441,13 @@ You can specify a cache directory in the ``pydap.lib.CACHE`` global variable. If
 Timeout
 ~~~~~~~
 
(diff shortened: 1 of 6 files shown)


jameshiebert:
This timeout option is definitely a good idea, but it looks like there are now some "interesting" conflicts in this PR now. I went through and *tried* to resolve them, but it might be safer if you take a shot at it (as you can more easily interpret your own intent.

laliberte:
@jameshiebert this one is also ready to merge. Should I go ahead?

Issue #98, opened 2017-05-10
Unable to parse token: <!DOCTYPE with HDF5 files


rvalenzuelar:
I'm having troubles when trying to open a hdf5 file:

```python
from pydap.client import open_url

In [2]: url_name='http://gpm1.gesdisc.eosdis.nasa.gov/opendap/GPM_L3/GPM_3IMERGH
   ...: HE.03/2016/105/3B-HHR-E.MS.MRG.3IMERG.20160414-S000000-E002959.0000.V03E
   ...: .HDF5'

In [3]: dataset = open_url(url_name)
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-3-854224b09f3a> in <module>()
----> 1 dataset = open_url(url_name)

/Users/raulvalenzuela/miniconda2/envs/grib_api/lib/python2.7/site-packages/pydap/client.pyc in open_url(url, application, session, output_grid)
     62     never retrieve coordinate axes.
     63     """
---> 64     dataset = DAPHandler(url, application, session, output_grid).dataset
     65 
     66     # attach server-side functions

/Users/raulvalenzuela/miniconda2/envs/grib_api/lib/python2.7/site-packages/pydap/handlers/dap.pyc in __init__(self, url, application, session, output_grid)
     62 
     63         # build the dataset from the DDS and add attributes from the DAS
---> 64         self.dataset = build_dataset(dds)
     65         add_attributes(self.dataset, parse_das(das))
     66 

/Users/raulvalenzuela/miniconda2/envs/grib_api/lib/python2.7/site-packages/pydap/parsers/dds.pyc in build_dataset(dds)
    159 def build_dataset(dds):
    160     """Return a dataset object from a DDS representation."""
--> 161     return DDSParser(dds).parse()
    162 
    163 

/Users/raulvalenzuela/miniconda2/envs/grib_api/lib/python2.7/site-packages/pydap/parsers/dds.pyc in parse(self)
     47         dataset = DatasetType('nameless')
     48 
---> 49         self.consume('dataset')
     50         self.consume('{')
     51         while not self.peek('}'):

/Users/raulvalenzuela/miniconda2/envs/grib_api/lib/python2.7/site-packages/pydap/parsers/dds.pyc in consume(self, regexp)
     39     def consume(self, regexp):
     40         """Consume and return a token."""
---> 41         token = super(DDSParser, self).consume(regexp)
     42         self.buffer = self.buffer.lstrip()
     43         return token

[...]

PR #99, opened 2017-05-10
Allow redirects even without credentials

laliberte:
This PR fixes issue #98. It solves it by allowing redirects even when no ``session`` is passed to the ``GET`` command. I've also added some testing to prevent regressions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ testing_extras = tests_require + [
     'pytest-cov',
     'pytest-attrib',
     'mock',
+    'requests-mock',
     'requests'
 ]
 
--- a/src/pydap/net.py
+++ b/src/pydap/net.py
@@ -60,10 +60,10 @@ def create_request(url, session=None):
         # be guaranteed to have all the needed credentials:
         return create_request_from_session(url, session)
     else:
-        # If a session object was not passed, we simply pass the requests
-        # library. The requests library allows the handling of redirects
-        # that are not naturally handled by Webob.
-        return create_request_from_session(url, requests)
+        # If a session object was not passed, we simply pass a new
+        # requests.Session() object. The requests library allows the
+        # handling of redirects that are not naturally handled by Webob.
+        return create_request_from_session(url, requests.Session())
 
 
 def create_request_from_session(url, session):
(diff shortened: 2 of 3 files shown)


src/pydap/net.py
```diff
@@ -56,24 +58,32 @@ def create_request(url, session=None):
...
-            # Missing schema can occur in tests when the url
-            # is not pointing to any resource. Simply pass.
-            pass
-    return Request.blank(url)
+        return create_request_from_session(url, session)
+    else:
+        # If a session object was not passed, we simply pass the requests
+        # library. The requests library allows the handling of redirects
+        # that are not naturally handled by Webob.
+        return create_request_from_session(url, requests)
```
jameshiebert: The ``requests`` module itself can be used as a session?! That's kind of bizarre and crazy.

setup.py
```diff
@@ -45,6 +45,7 @@
     'pytest-cov',
     'pytest-attrib',
     'mock',
+    'requests-mock',
```
jameshiebert: Does this get used?

src/pydap/net.py
```diff
@@ -23,7 +24,8 @@ def GET(url, application=None, session=None):
 
 
 def raise_for_status(response):
-    if response.status_code >= 400:
+    # Raise error if status is above 300:
+    if response.status_code >= 300:
```
jameshiebert: What is the effect of this? If the client gets a redirect, this will raise and exception, but then what handles it?

laliberte: The client already attempts (through the ``session.head`` call) to follow redirects. It is fails then an error should be rightly raised. At least that's the thinking...

src/pydap/net.py
laliberte: The ``requests`` module has almost all the same methods as the ``session`` object, so yes, it can be used instead of a ``session`` it our context. I used this approach because it seemed the cleanest but I can see that you disagree :) The alternative would be to replace ``requests`` by ``requests.Session()``, i.e. create a new temporary ``session``.

setup.py
laliberte: It does, in the tests. But it seems that I forgot to include in the tests when I pushed. I've included them now.

laliberte:
Tests have passed, I'm merging. I've fixed some of things that you found confusing.

PR #102, opened 2017-05-20
Fix CEDA openid authentication

laliberte:
This PR fixes the authentication scheme to ensure that CEDA openids are working. It also includes a test to prevent regressions.

This fix replaces the ``mechanicalsoup`` dependency with ``beautifulsoup4``. I think this is for the best since ``beautifulsoup4`` is a generally better maintained package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,4 +11,4 @@ install:
 
 script:
     - flake8
-    - py.test -a 'not auth' --cov=src/pydap/
+    - py.test --cov=src/pydap/
--- a/docs/client.rst
+++ b/docs/client.rst
@@ -350,8 +350,9 @@ Authentication is done through an ``openid`` and a ``password``:
 
     >>> from pydap.client import open_url  
     >>> from pydap.cas.esgf import setup_session 
-    >>> session = setup_session(openid, password)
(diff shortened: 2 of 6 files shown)


laliberte:
This fixes #103. @jameshiebert could you review next week?

jameshiebert:
Do you have a run on Travis with the auth switch turned on (and all of the CEDA environment variables set)? I mean, I trust that this works, but it'd be good to confirm.

laliberte:
Give me a few hours and I'll do this!

Fred

laliberte:
@jameshiebert I've converted CAS tests to ``pytest``. These tests are now skipped if the required environment variables are not set. They can still be skipped with ``py.test -a 'not auth'``. This way, a contributor (e.g. me) can set their travis environment to pass those tests but it does not require the main repository to pass them. This means that in the future when a PR passes my Travis then these tests will be successful.

jameshiebert:
Yeah, that's cool the way that you did that! Unfortunately (or... fortunately that we caught it?)... these [tests failed on your TravisCI](https://travis-ci.org/laliberte/pydap/builds/234996386)...

laliberte:
Darn!

PR #90, opened 2017-04-05
Remove scipy from travis-ci to close #84

laliberte:
This PR completely removes ``scipy`` from testing following the publication of version 3.0.6 of ``gsw`` on [PyPI](https://github.com/TEOS-10/python-gsw/issues/41). This speeds up our builds significantly, hurray! Also, it closes #84.

PR #105, opened 2017-05-22
Hot fix to ensure that CEDA openid fix is working

laliberte:
This PR fixes #102. It switches ``html5lib`` to ``lxml`` parser and adds ``lxml`` as a dependency in ``setup.py``. This PR also removes ``scipy`` from dependencies. It must have been re-added by mistake (#90).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ install_requires = [
     'Jinja2',
     'docopt',
     'six >= 1.4.0',
-    'beautifulsoup4',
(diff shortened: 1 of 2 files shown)


laliberte:
@jameshiebert this PR now works. It passes tests on the main and python 3.4-3.6 on my travis. For some reason my travis does not have the same setting as the main and it does not catch warnings properly... puzzling. For our purposes, however, it should OK.

PR #106, opened 2017-05-22
Add a netcdf4-python interface for testing

laliberte:
This PR uses the devel server to launch a temporary opendap service that can be queried by ``netcdf4-python``. This will be particularly useful for testing the netcdf4 handler (enable roundtrips). Also, this enables users to query opendap urls served by CAS (ESGF, URS) by still have the same interface as ``netcdf4-python``. This is mostly for convenience but since it is fairly low maintenance I think this convenience should be provided to users of pydap.


laliberte:
@jameshiebert any thoughts about this one?

laliberte:
@laliberte this can now be merged

PR #109, opened 2017-05-23
Update README.md

danbgray:
The readme should include some of the basic info from the main docs to help people get set up without digging too much deeper.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,69 @@ Pydap
 [![documentation](https://readthedocs.org/projects/pydap/badge/?version=latest)](http://pydap.readthedocs.org/en/latest/)
 [![PyPI](https://img.shields.io/pypi/v/pydap.svg?maxAge=2592000?style=plastic)](https://pypi.python.org/pypi/Pydap/)
 
-[Pydap](http://pydap.readthedocs.io/en/latest/) is an implementation of the Opendap/DODS protocol, written from scratch. 
-You can use Pydap to access scientific data on the internet without having to 
+[Pydap](http://pydap.readthedocs.io/en/latest/) is an implementation of the Opendap/DODS protocol, written from scratch in pure python.  You can use Pydap to access scientific data on the internet without having to 
 download it; instead, you work with special array and iterable objects that 
 download data on-the-fly as necessary, saving bandwidth and time. The module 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,9 +7,69 @@ Pydap
...
+              -1.41666666e-01],
+            [  6.38000011e-01,   8.95384610e-01,   7.21666634e-01,
+               8.10000002e-01]]], dtype=float32), array([ 366.]), array([-69., -67., -65., -63.]), array([ 41.,  43.,  45.,  47.])]
+```
+
+For more information, please check the documentation on [using Pydap as a client](http://pydap.readthedocs.io/en/latest/server.html). Pydap also comes with a simple server, implemented as a [WSGI]( http://wsgi.org/) application. To use it, you first need to install a data handler:
+
+```bash
+
+    $ pip install pydap.handlers.netcdf
```
jameshiebert: Add ``pip install Pydap[server]`` here

README.md
```diff
@@ -7,9 +7,69 @@ Pydap
...
+```
+This will install support for [netCDF](http://www.unidata.ucar.edu/software/netcdf/) files; more [handlers](http://pydap.readthedocs.io/en/latest/handlers.html) for different formats are available, if necessary. Now create a directory for your server data:
+
+```bash
+
+    $ paster create -t pydap myserver
```
jameshiebert: This part can be omitted

README.md
```diff
@@ -7,9 +7,69 @@ Pydap
...
+```
+To run the server just issue the command:
+
+```bash
+
+    $ paster serve ./myserver/server.ini
```
jameshiebert: Replace this with ``$ pydap --data ./myserver/data/ --port 8001``

jameshiebert:
Looks great. Thanks!

danbgray:
One caveat:
the current version of pydap.handlers.netcdf on pypi is 0.6.4 which is not compatible with python3,/ had an error on pip install about a print statement not having parentheses. The version 0.7 that is currently published works fine

jameshiebert:
Oh right. We actually merged the ``pydap.handlers.netcdf`` into the mainline repo (as an extra dependency) in #35, so I'll update the docs on that. Thanks for pointing that out.

PR #111, opened 2017-05-24
move src/pydap to pydap

tomkralidis:
Pull request to `mv src/pydap pydap/`.


laliberte:
While you're reorganizing things around, maybe we could give some thoughts to moving most the files to a ``pydap/dap2`` subdirectory and create an empty ``pydap/dap4`` subdirectory. I don't think we will be implementing dap4 soon but it would be nice to have a place holder in case someone wants to start working on it.

What are your thoughts @jameshiebert ?

tomkralidis:
@laliberte thanks for the info.  I'm guessing `dap2` and `dap4` are versions of OPeNDAP?  In which case we can probably go with `pydap/dap/dap2` and `pydap/dap/dap4` type thing?

laliberte:
I'm not sure here. That's why I wanted to see what @jameshiebert was thinking. Some part of ``pydap`` could most likely be used for both ``dap2`` and ``dap4``.  Maybe we could simply create empty ``pydap/dap2`` and  ``pydap/dap4`` subdirs for the moment.

jameshiebert:
I'm extremely hesitant to move forward with moving the entire source tree without a very good reason. As long as the `setup.py` file points to the source files, [which it does](https://github.com/pydap/pydap/blob/master/setup.py#L82), AFAIK it doesn't really matter where the source files are.

Making a big move like this will make it very challenging to use diff tools to compare anything before this change with anything after. And any outstanding PRs will require a more difficult rebase.

Could you describe more of you motivation or explain how the advantages of such a move would outweigh the disadvantages?

tomkralidis:
@jameshiebert thanks for the feedback.  The motivation is mostly myself not having to navigate an extra (always empty) directory as a developer working on Pydap :).  Most Python projects (that I have worked on) usually place the package at the top level without an empty 'src' directory in between.  Perhaps there is some rationale for the `src/pydap` directory which I'm not aware of?

In summary, this PR is not to address a concrete bug, but introduces cosmetic changes aligning with [PyPUG](https://packaging.python.org/) (sample setup [here](https://github.com/pypa/sampleproject)).  Having said this, I'm happy to close this PR if there's no interest or causes too much heartache.

laliberte:
I agree that eventually, ``pydap`` should make this move. However, I'm not sure version 3.2.2 is the right version. Maybe wait until 3.3 ?

jameshiebert:
@tomkralidis Yeah, I agree that it's non-ideal to have an empty directory in the middle of the path, and maybe there was some reason historically (that pre-dates my tenure of working on Pydap). But overall, I think that it's a pretty minor inconvenience that can be overcome through other steps in our workflow (e.g. we could just symlink `pydap` -> `src/pydap`).

I've had other projects where we have done something like this, and we basically lost the ability to (easily) do source tree comparisons from before to after.

Though I appreciate the spirit of this PR, I'm inclined to close is for those reasons. Any strong objections?

jameshiebert:
PS It's *possible* that the fact that `Pydap` has [multiple namespace packages](https://github.com/pydap/pydap/blob/master/setup.py#L85) under the `src` tree affects the need for this, but I'm not sure.

tomkralidis:
@jameshiebert thanks for the feedback and consideration. +1 to close.

PR #115, opened 2017-06-02
Remove non-existent entry in index

ghisvail:


laliberte:
The Changelog is not used at the moment but it should probably point to ``NEWS.md``. @ghisvail could see whether that's possible to do?

laliberte:
@ghisvail have you had the chance to take a look?

ghisvail:
I was busy dealing with the update of the Debian packaging. No idea whether Sphinx will be happing with Markdown input. Feel free to merge this and try out by yourself afterwards.

Issue #119, opened 2017-06-04
LocalTestServer without external dependencies for server side functions


shoyer:
I just experimented a little with using Pydap's new `LocalTestServer` feature for xarray. This looks great, but using it requires server side functions, which adds two additional dependencies for pydap -- gsw and coords.

For a local test setup (e.g., for verifying that pydap is working correctly with xarray), ideally we could make things work with just pydap.

rabernat:
I just started playing with pydap and hit this same issue.

It would be nice to have an easy way to bypass these server side functions from the command line, or simply not apply them if the dependencies are not available.

```
Traceback (most recent call last):
  File "/Users/rpa/Code/pydap/src/pydap/lib.py", line 286, in load_from_entry_point_relative
    globals(), None, [r.attrs[0]], 1),
  File "/Users/rpa/Code/pydap/src/pydap/wsgi/functions.py", line 28, in <module>
    import gsw
ModuleNotFoundError: No module named 'gsw'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 135, in handle
    self.handle_request(listener, req, client, addr)
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 176, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/webob/dec.py", line 131, in __call__
    resp = self.call_func(req, *args, **self.kwargs)
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/webob/dec.py", line 196, in call_func
    return self.func(req, *args, **kwargs)
  File "/Users/rpa/Code/pydap/src/pydap/wsgi/app.py", line 184, in __call__
    return req.get_response(self.app)
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/webob/request.py", line 1327, in send
    application, catch_exc_info=False)
  File "/Users/rpa/miniconda3/envs/pangeo_pydap/lib/python3.6/site-packages/webob/request.py", line 1291, in call_application
    app_iter = application(self.environ, start_response)
[...]

PR #123, opened 2017-06-05
Remove serverside functions in devel server (solves #119)

laliberte:
This PR solves #119. Server side functions can still be tested with the development server but will require an extra step from the user.


rabernat:
This PR is almost a year old. Is there any chance it will be merged soon?

Issue #120, opened 2017-06-04
add Requests to install_requires


wholmgren:
The requests package is listed under ``cas_extras`` instead of ``install_requires``. ``pydap.client`` needs ``requests``, so it seems to me that it should be a standard dependency.

PR #124, opened 2017-06-05
Explicitly test client-only installations (solves #120)

laliberte:
This PR restructure tests so that a pure client installation is always tested in Travis. The proposed approach can be easily extended to other extras. This should prevent future regressions like #120.


tomkralidis:
@laliberte can you take a look at merge conflicts here?  Thanks again.

laliberte:
@tomkralidis I think this one is ready

PR #125, opened 2017-06-05
add failing test

ocefpaf:
This PR adds a test that passes with `pydap 3.2.1` but fails with `pydap 3.2.2`. Not sure what caused the regression and there is no need to merge this PR if the devs don't want to test against a live server that may not be always available. I only sent this to make it easier to debug in case someone have the time to bisect the last few PRs.


laliberte:
This was fixed in #127.

Issue #126, opened 2017-06-05
open_url does not handle gzip encoding


laliberte:
It appears that pydap does not handle gzip encoding. This is why an error is raised by #125.

PR #127, opened 2017-06-05
Handle gzipped responses (Solves #126)

laliberte:
This PR allows the handling of gzipped responses. In doing so, it introduces the possibility of providing a gzipped response in ``BaseHandler``.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -7,6 +7,7 @@ dataset to the internal model.
 """
 
 import io
+import gzip
 import sys
 import pprint
 import copy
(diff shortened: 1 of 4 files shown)


laliberte:
I will also merge this one to get rid of some of the backlog. It is more comprehensive than #152 (it includes tests and server gzip capabilities), so it will be the PR that gets merged. The approach in #152 is interesting in its simplicity. Maybe someone could look whether this implementation could be simplified.

ocefpaf:
It would be nice to get a new release, fixes #127 that still plagues some of us :grimacing:

Issue #133, opened 2017-07-24
pydap.__version__


shoyer:
It would be nice to have this information available at runtime (e.g., for https://github.com/pydata/xarray/pull/1485).

tomkralidis:
Associated pull request in #135

tomkralidis:
Fixed in master.

PR #135, opened 2017-09-16
add pydap.__version__ capability (#133)

tomkralidis:
As per #133


src/pydap/__init__.py
```diff
@@ -3,3 +3,5 @@
 '''
 
 __import__('pkg_resources').declare_namespace(__name__)
+
+__version__ = __import__('pkg_resources').get_distribution(__name__).version
```
ocefpaf: I'm not a fan of `pkg_resources`, it is slow and hackish.
Could we try something like versioneer? Or maybe define the version in `__init__.py` and parse it in `setup.py`?

tomkralidis:
@ocefpaf fixed.  Travis errors (Python 3.5) are unrelated. This PR requires squashing when merging.

ocefpaf:
LGTM. @jameshiebert this one is an easy merge :wink:

jhamman:
this looks good. If you are looking for a standard (likely more robust) approach, you may want to checkout [versioneer](https://github.com/warner/python-versioneer).

tomkralidis:
Thanks @jhamman. versioneer looks great, but I'd prefer to not include another dependency.

jhamman:
Okay, for clarity sake though, versioneer is not an external dependency.

ocefpaf:
:+1: for `versioneer`, it is a one time thing and you never have to worry about versions again :stuck_out_tongue_winking_eye:

Issue #136, opened 2017-09-17
add support for THREDDS Catalog XML


tomkralidis:
Implement enhancement to enable a THREDDS Catalog XML representation available at each directory level.

Example link (see bottom of page before footer): http://test.opendap.org/opendap/conus/
Example XML: http://test.opendap.org/opendap/conus/catalog.xml

tomkralidis:
Implemented in #138

PR #138, opened 2017-09-17
add support for THREDDS Catalog XML (#136)

tomkralidis:
This PR implements #136.

### DIFF
--- a/src/pydap/tests/test_wsgi_app.py
+++ b/src/pydap/tests/test_wsgi_app.py
@@ -3,6 +3,7 @@
 import os
 import tempfile
 import shutil
+from xml.etree import ElementTree as etree
 
 from webtest import AppError
 from webtest import TestApp as App
@@ -69,6 +70,14 @@ class TestDapServer(unittest.TestCase):
         """Test invalid DAP requests."""
         with self.assertRaises(ExtensionNotSupportedError):
             self.app.get("/README.txt.dds")
+
+    def test_thredds_catalog_request(self):
+        """Test that THREDDS Catalog requests work."""
+        res = self.app.get("/catalog.xml")
+        self.assertEqual(res.status, "200 OK")
+        self.assertTrue(res.text.startswith('<?xml'))
+        xml = etree.fromstring(res.text)
+        self.assertEqual(xml.tag, '{http://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0}catalog')  # noqa
 
     def test_not_found(self):
         """Test 404 responses."""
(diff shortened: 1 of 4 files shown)


neishm:
This is a very nice feature to have in pydap.  One minor problem I encountered while testing - catalogs in subdirectories were not giving valid URLs.  I found that removing the ```{{ location }}}``` from the ```urlPath``` attributes in ```wsgi/templates/catalog.xml``` seemed to fix that problem for me.

tomkralidis:
@neishm can you provide an example of what you're expecting?

neishm:
For instance, if I have a file **abc** in directory **xyz**, then navigating to `http://localhost:8001/abc/catalog.xml` gives me
```xml
<?xml version="1.0" ?>
<!-- Pydap 3.2.2 -->
<catalog xmlns="http://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://www.unidata.ucar.edu/namespaces/thredds/InvCatalog/v1.0 http://www.unidata.ucar.edu/schemas/thredds/InvCatalog.1.0.6.xsd">
    <service name="odap" serviceType="OPeNDAP" base="/xyz/"/>
    <service name="file" serviceType="HTTPServer" base="/xyz/"/>
    <dataset name="abc " ID="/xyz/abc">
        <dataSize units="bytes">333189232</dataSize>
        <date type="modified">2017-02-15T13:13:15.148454</date>
        <access serviceName="odap" urlPath="/xyz/abc"/>
        <access serviceName="file" urlPath="/xyz/abc"/>
    </dataset>
```

If I open this catalog in panoply or IDV, and try to access **abc** from the catalog I get a 404 error.  The reason is it's trying to access `http://localhost:8001/xyz//xyz/abc` (directory name repeated).

Checking the [THREDDS docs](http://www.unidata.ucar.edu/software/thredds/current/tds/tutorial/CatalogPrimer.html#Constructing_an_access_URL), looks like it's combining the **base**+ **urlPath** to get the final URL on the client side.  So, I think either **base** should just be "/", or the **urlPath** should exclude the directory structure (e.g., just be "abc" in this example).  I tested the latter approach, and that resolved the issue for me.

Would really like to have THREDDS catalog support for pydap.  Already testing out your branch on our local server, and it makes things a lot easier to navigate our datasets remotely.

tomkralidis:
@neishm thanks for the info.  I'll look into this tonight. The important item is validating this against the specification, and thanks for pointing to the THREDDS docs.

Aside: can you contact me for an offline question? tom.kralidis@gmail.com

Issue #139, opened 2017-09-18
Handlers for files without extensions


neishm:
Some file types do not use a standard file extension to identify them.  They may have to be opened to inspect the first few bytes, etc.  Is it possible to support this within Pydap? Perhaps, in addition to providing ```extensions```, Handlers could implement a method to determine if the file is supported?  E.g., something like

```python
from pydap.handlers.lib import BaseHandler
class My_Handler(BaseHandler):
  @staticmethod
  def can_handle (filepath):
    # Check file contents (for "magic patterns", etc.)
    with open(filepath,'rb') as f:
      if some_criteria(f): return True
      else: return False
  # ...
```

Then ```pydap.handlers.lib.get_handler``` could query the ```can_handle``` routines when no extensions match the file?

PR #140, opened 2017-09-18
Allow Handlers to provide their own criteria for handling the filepat…

neishm:
…h. (#139)


tomkralidis:
@neishm thanks.  Is `can_handle` a documented handler function?  Should it be?

neishm:
That's a good point.  I'll add some documentation for it in `docs/developer_handlers.rst`.  I'm also open for alternative suggestions on how to choose the appropriate handler for the files.  For instance, Iris has a more sophisticated `FormatAgent` structure which allows you to choose either file extensions or magic byte strings to determine the handler ([source](https://github.com/SciTools/iris/blob/7d0cced495838acc845986f41eb347886117f278/lib/iris/fileformats/__init__.py)).  That might be overkill, though...

neishm:
Some documentation added, still need to create some tests.

captainkirk99:
Is this PR still active?

neishm:
I no longer need this functionality myself.  If nobody else needs it, could be closed without merging.

PR #143, opened 2017-09-19
add LICENSE file (#142)

tomkralidis:


laliberte:
This is also simple maintenance. I will merge.

PR #149, opened 2017-10-09
change encoding of NaN values (#148)

### LINKED ISSUE #148: Issues with encoding of NaN
It seems that the NetCDF-Java OPeNDAP client has a problem with pydap's encoding of NaN (e.g. in the DAS response). Some clients seem to only parse 'NaN', while pydap encodes it as 'nan'.

The DAP2 specification does not mention NaN at all, IEEE754 does not specify any string representation of NaN. http://www.unidata.ucar.edu/mailing_lists/archives/thredds/2016/msg00145.html
According to the link above it looks like THREDDS changed to using NaN.

pauleckhardt:
fixes #148

### DIFF
--- a/src/pydap/lib.py
+++ b/src/pydap/lib.py
@@ -127,7 +127,7 @@ def encode(obj):
     try:
         if isnan(obj):
             return "NaN"
-    except TypeError:
+    except:
         pass
     try:
         return '%.6g' % obj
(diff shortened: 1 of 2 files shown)


pauleckhardt:
Messed up by using my master branch as a target. Will open a new pull request.

Issue #150, opened 2017-11-10
Error reading data GEOS-5


nqlong:
Hi everyone, I got the problem when using pydap get data from GEOS-5 ([https://opendap.nccs.nasa.gov/dods/GEOS-5/fp/0.25_deg/fcast/inst1_2d_hwl_Nx])

I write the code as the guide:
```
dataset_geos= open_url('https://opendap.nccs.nasa.gov/dods/GEOS-5/fp/0.25_deg/fcast/inst1_2d_hwl_Nx/inst1_2d_hwl_Nx.20171110_00')
so2cmass = dataset_geos['so2cmass']
so2cmass_dt=so2cmass[0,10:14,10:14]
print(so2cmass_dt.data)
```

and it return the error

>  Traceback (most recent call last):
>   File "C:\Program Files (x86)\python_3_6_1\lib\site-packages\pydap\model.py", line 404, in _getitem_string
>     return self._dict[quote(key)]
> KeyError: 'so2cmass%2Eso2cmass'
> 
> During handling of the above exception, another exception occurred:
> 
> Traceback (most recent call last):
>   File "C:\Program Files\JetBrains\PyCharm 2017.1.1\helpers\pydev\pydevd.py", line 1578, in <module>
>     globals = debugger.run(setup['file'], None, None, is_module)
>   File "C:\Program Files\JetBrains\PyCharm 2017.1.1\helpers\pydev\pydevd.py", line 1015, in run
>     pydev_imports.execfile(file, globals, locals)  # execute the script
>   File "C:\Program Files\JetBrains\PyCharm 2017.1.1\helpers\pydev\_pydev_imps\_pydev_execfile.py", line 18, in execfile
>     exec(compile(contents+"\n", file, 'exec'), glob, loc)
>   File "E:/Project/open_air/tests/geos-5.py", line 12, in <module>
>     so2cmass_dt = so2cmass[0,10:14,10:14]
>   File "C:\Program Files (x86)\python_3_6_1\lib\site-packages\pydap\model.py", line 740, in __getitem__
>     var.data = self[var.name].data[slice_]
>   File "C:\Program Files (x86)\python_3_6_1\lib\site-packages\pydap\handlers\dap.py", line 149, in __getitem__
>     return dataset[self.id].data
>   File "C:\Program Files (x86)\python_3_6_1\lib\site-packages\pydap\model.py", line 425, in __getitem__
>     return self._getitem_string(key)
[...]

PR #151, opened 2017-11-10
Trap IndexErrors in StructureType._getitem_string, when attempting to apply a split string.

neishm:
Seems to fix issue #150, hopefully for the right reasons.  I'm not an expert on the code base, so a review by a more knowledgeable person would be appreciated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,7 +45,7 @@ tests_require = (functions_extras + cas_extras + server_extras +
                  hdl_netcdf_extras +
                  ['WebTest',
                   'beautifulsoup4',
-                  'flake8'])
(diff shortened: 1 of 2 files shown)


neishm:
I would just like to mention that the Travis CI build is "failing" because the source-code checker (flake8) is complaining about "bare except" statements in lines I never touched.

This appears to stem from a recent update of flake8 (version 3.5.0, released 2017-10-23).  Downgrading to the previous version of flake8 (3.4.1) causes all source code checks to pass.

laliberte:
I merged PR #159 that solves this issue. I did not see your PR, sorry!

neishm:
No problem!

PR #152, opened 2017-11-16
Fix gzip compression (#126)

regeirk:
I had o problem loading data from the Global Forecast System (GFS). It raised a UnicodeDecodeError exception, which after a while I figured out it happens because the data is gzipped. I came up with a solution modifying the `GET` function in the `net` module, adding `response.decode_content()`.

### DIFF
--- a/src/pydap/net.py
+++ b/src/pydap/net.py
@@ -23,8 +23,11 @@ def GET(url, application=None, session=None, timeout=DEFAULT_TIMEOUT):
         _, _, path, query, fragment = urlsplit(url)
         url = urlunsplit(('', '', path, query, fragment))
 
-    return follow_redirect(url, application=application, session=session,
-                           timeout=timeout)
+    response = follow_redirect(url, application=application, session=session,
+                               timeout=timeout)
+    # Decode request response (i.e. gzip)
+    response.decode_content()
+    return response
 
 
 def raise_for_status(response):


MiCurry:
+1ing this one. Had trouble accessing a GrADS server (http://nomads.ncep.noaa.gov:9090/dods/rtofs/). This fixed work, thanks @regeirk.

ForTozs:
I tried these changes and it stopped the UnicodeDecodeError for me as well. I should note, that the behavior is slightly changed. Before I could call a datasource with:

```
>>> latitudes = meta['rows'][:]
>>> latitudes
array([44.88375, 44.87625, 44.86875, ..., -0.09375, -0.10125, -0.10875])
```

This would result in latitudes being an array datatype. Now, the datatype has changed to BaseType with data array. If I want the data array, now it is necessary to call latitudes as:

```
>>> latitudes = meta['rows'][:]
>>> latitudes
<BaseType with data array([44.88375, 44.87625, 44.86875, ..., -0.09375, -0.10125, -0.10875])>
>>>latitudes.data
array([44.88375, 44.87625, 44.86875, ..., -0.09375, -0.10125, -0.10875])
```

Sorry if this is convoluted or off-topic. Just thought I'd share how it changed the behavior of my code in case anyone else comes across this. It may be that this happened during the update and is unrelated to this previous bug.

ocefpaf:
@tomkralidis it would be nice if this one got merged and a new issue releases.

PR #156, opened 2017-12-08
bugfix: unicode attribute values and variable values (#155)

### LINKED ISSUE #155: problems with unicode attribute values and variable values
When using a unicode string as attribute value or variable value, a UnicodeEncodeError is thrown in
pydap/lib.py", line 135, in encode:
    return '"{0}"'.format(obj)

pauleckhardt:
Fixes #155

### DIFF
--- a/src/pydap/lib.py
+++ b/src/pydap/lib.py
@@ -125,10 +125,10 @@ def quote_val(name):
        Used for attributes and values.
        Encode every cahracter which is not 7bis ASCII.
 
-        >>> quote("White space. /;")
+        >>> quote_val("White space. /;")
         'White space. /;'
-        >>> quote("<Ue>mlaut")
-        '%C3%9Cmlaut'
+        >>> quote_val("Period.")
+        'Period.'
 
     """
     safe = ''.join([chr(x) for x in range(128)])
--- a/src/pydap/lib.py
+++ b/src/pydap/lib.py
@@ -8,7 +8,6 @@ from six import binary_type, MAXSIZE
 
 from .exceptions import ConstraintExpressionError
 
-
 __dap__ = '2.15'
 __version__ = get_distribution("Pydap").version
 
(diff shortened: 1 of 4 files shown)


src/pydap/lib.py
```diff
@@ -120,14 +119,30 @@ def quote(name):
...
+
+    """
+    safe = ''.join([chr(x) for x in range(128)])
+    return quote_(name.encode('utf-8'), safe=safe)
 
 def encode(obj):
     """Return an object encoded to its DAP representation."""
+    try:
+        return quote_val(obj)
+    except:
```
laliberte: Which exception is expected here? You should probably except only that one here.

src/pydap/responses/dods.py
```diff
@@ -203,18 +204,26 @@ def _basetype(var):
...
-                        raise TypeError("Could not convert word '{0}' to bytes"
-                                        .format(word))
+                    try:
+                        enc = quote_val(word)
+                        length = len(enc)
+                        yield tostring_with_byteorder(
+                                    np.array(length),
+                                    np.dtype(DAP2_ARRAY_LENGTH_NUMPY_TYPE))
+                        yield enc
+                    except:
```
laliberte: Again here, what is your try clause capturing? It is too wide and "yield tostring_wit..." is repeated. This must be avoided.

src/pydap/lib.py
tomkralidis: Echo'ing @laliberte, @pauleckhardt can we be more explicit here?

src/pydap/responses/dods.py
tomkralidis: Echo @laliberte, need to be more explicit here.

captainkirk99:
This is a very old PR. Is it still active?

Mikejmnez:
Is this PR active?

Mikejmnez:
Closing this issue as it is very old

PR #161, opened 2018-02-06
Fix bad link in README

flackdl:
The README linked to http://pydap.readthedocs.io/en/latest/server.html when it should have linked to http://pydap.readthedocs.io/en/latest/client.html.


jameshiebert:
Ha, good catch! Thanks.

PR #173, opened 2018-10-13
Fix unit tests in Python 3.5

betodealmeida:
Unit tests were failing in Python 3.5 because for some reason `encode(numpy.array('1', dtype='<U1'))` was returning `1` instead of `"1"`. I fixed it with an explicit check, though it would be good to refactor the `encode` function to make it more robust.

I also fixed a deprecation warning, using `frombuffer` instead of `fromstring`.


7yl4r:
A weird quirk, but it looks to me like this workaround will have to do. Might consider adding a comment in `encode` to note the reason for the inf/else is py 3.5 compatibility.

PR #179, opened 2018-11-29
add netcdf handler entry point

### LINKED ISSUE #129: NetCDF handler not registered
When I do a fresh install of pydap (python3 setup.py install --user) the built-in NetCDF handler does not seem to be enabled. I believe an entry point for it in setup.py is needed:

```
$ git diff setup.py
diff --git a/setup.py b/setup.py
index 7471f20..ba25bd1 100644
--- a/setup.py
+++ b/setup.py
@@ -97,6 +97,9 @@ setup(name='Pydap',
       },
       test_suite="pydap.tests",
       entry_points="""
+            [pydap.handler]
+            nc = pydap.handlers.netcdf:NetCDFHandler
+
             [pydap.response]
             das = pydap.responses.das:DASResponse
             dds = pydap.responses.dds:DDSResponse
```

At least it works in my case after doing that. The same procedure probably applies for the CSV handler.

Best regards,
Jesper

### LINKED ISSUE #165: can't get netcdf handler to work
I am just trying to run a pydap server in its most simple configuration, to serve netcdf files.

I installed pydap as follows:
```
pip install Pydap[server,handlers.netcdf]
```

Then I ran a pydap server from a directory with netcdf files in it
```
pydap --data .
```

The netcdf (.nc) files are visible in the browser but not active. I can't click them.

The only way I can get them to work is by cloning https://github.com/pydap/pydap.handlers.netcdf and installing that package. But that package hasn't been updated in more than 5 years.

My understanding is that this package (the main pydap package) provides the netcdf handler. I just can't figure out how to properly install the handler.

I would appreciate any advice anyone might have.

rabernat:
This is the simple fix recommended by @jblarsen  in https://github.com/pydap/pydap/issues/129#issue-238923262.

fixes #129
fixes #165
 
Basically, it seems like the netcdf handler entry point was not being installed by `setup.py`. So when `load_handlers()` gets called (https://github.com/pydap/pydap/blob/master/src/pydap/handlers/lib.py#L43), no netcdf handler was being found. Consequently, when I tried to run a server
```
pydap --data /path/with/netcdf/files
```
The netcdf files were not being handled correctly.

Contrary to what the [docs say](http://pydap.org/en/latest/handlers.html#installing-data-handlers), the following does not seem to properly set up the entry point:
```
$ pip install Pydap[handlers.netcdf]
```
I just get
```
Requirement already satisfied: Pydap[handlers.netcdf] in ./src (3.2.2)
```

Evidently this situation is not covered by the test suite. I wonder if an integration test involving running the command-line app would be useful.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -123,6 +123,9 @@ setup(name='pydap',
       },
       test_suite="pydap.tests",
       entry_points="""
+            [pydap.handler]
+            nc = pydap.handlers.netcdf:NetCDFHandler
+
             [pydap.response]
             das = pydap.responses.das:DASResponse
             dds = pydap.responses.dds:DDSResponse


rabernat:
Any of the pydap devs have any thoughts on this? @tomkralidis?

tomkralidis:
I'm good with merging.  As for tests running the command line app, if this is to validate that the command line app runs properly, can/should we drop to Travis to test?

rabernat:
> As for tests running the command line app, if this is to validate that the command line app runs properly, can/should we drop to Travis to test?

I don't have enough experience with testing to know the best approach here. I think that can be handled in a future PR.

rabernat:
Is there anything else I can do on this PR to move it closer to acceptance?

flackdl:
I was running into the same issue.  I noticed the pypi pyap version hasn't been updated in a while (2017).  

Is it recommended to install this manually via pip + git and all the relevant dependencies?

tomkralidis:
@flackdl probably a good idea until we issue a new release, yes.

PR #180, opened 2019-01-02
Encode query string before performing the request

crossi202:
Please review this pull request.
It would encode the query string in the URL, to prevent the issue described in https://github.com/Unidata/thredds/issues/1144#issuecomment-420107544 and https://www.unidata.ucar.edu/support/help/MailArchives/netcdf/msg14389.html

Indeed, some servers, like Apache Tomcat 8.5.33, do not accept the chars `[` and `]`, if not encoded.


captainkirk99:
Is this PR still active?

captainkirk99:
This is failing CI like this:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/pydap/model.py:323: in __getitem__
    out.data = self._get_data_index(index)
src/pydap/model.py:353: in _get_data_index
    return self._data[index]
src/pydap/handlers/dap.py:175: in __getitem__
    dataset.data = unpack_data(BytesReader(data), dataset)
src/pydap/handlers/dap.py:452: in unpack_data
    return unpack_children(xdr_stream, dataset)
src/pydap/handlers/dap.py:383: in unpack_children
    out.append(tuple(unpack_children(stream, col)))
src/pydap/handlers/dap.py:383: in unpack_children
    out.append(tuple(unpack_children(stream, col)))
E   RecursionError: maximum recursion depth exceeded in comparison
!!! Recursion detected (same locals & position)
____________________________ TestPeriod.test_client ____________________________

self = <pydap.tests.test_special_chars.TestPeriod testMethod=test_client>

    def test_client(self):
        dataset = open_url("http://localhost:8001/", application=self.app)
        self.assertEqual(dataset["a.b"].name, "a%2Eb")
>       self.assertEqual(dataset["a.b"][0], 1)

```

jgallagher59701:
This is an important PR and we should run the issue to ground. Sorry I cannot be more helpful with the actual problem, but this will break the client for virtually all hyrax and TDS instances that use Tomcat 8.

src/pydap/net.py
```diff
@@ -3,6 +3,10 @@
 from contextlib import closing
 import requests
 import ssl
+try:
+    from urllib import quote  # Python 2.x
```
owenlittlejohns, line 7: @jgallagher59701 and I had a quick chat, and I think we can remove this optional import for Python 2 support. The PyDAP CI/CD is only testing against Python 3 (3.7 or newer), and Python 2 sunset date was the 1st of January 2020.

src/pydap/net.py
```diff
@@ -25,6 +29,10 @@ def GET(url, application=None, session=None, timeout=DEFAULT_TIMEOUT,
         _, _, path, query, fragment = urlsplit(url)
         url = urlunsplit(('', '', path, query, fragment))
 
+    scheme, host, path, query, fragment = urlsplit(url)
+    query = quote(query)
+    url = urlunsplit((scheme, host, path, query, fragment))
```
owenlittlejohns, line 34: I think there should be a unit test or two added to ensure the URL query gets encoded as expected. The things that I think would make a good test:
* Seeing a URL with the `[` and `]` characters in the query (so I'm guessing a constraint expression).
* Making sure slashes in variable names are handled correctly (I _think_ `quote` may consider `/` to be a safe character).
* Making sure the rest of the URL is not encoded.

I think one way to do this test would potentially be to make the `follow_redirect` function and then look at the value of the `url` argument.

owenlittlejohns (changes requested):
@crossi202 - I think there are a couple of quick things to tweak in this PR, as called out on the file itself. Because this PR has been open a while, I'll wait and see if you wanted to make those changes. If not, I can try and pick them up.

jjrob:
Thank you devs for pydap. I have used it for a long time but only just recently I started working with it on https://www.earthdata.nasa.gov/, into which a very large amount of physical oceanographic data has migrated from https://podaac.jpl.nasa.gov/ over the past few years. I can confirm the issue raised by @crossi202 exists there with pydap 3.4.0:

> Indeed, some servers, like Apache Tomcat 8.5.33, do not accept the chars `[` and `]`, if not encoded.

If anyone needs to get this working, a quick fix is to patch net.py to change this:
```{python}
    response = follow_redirect(url, application=application, session=session,
                               timeout=timeout, verify=verify)
```
to this:
```{python}
    response = follow_redirect(url.replace('[','%5B').replace(']','%5D'), application=application, session=session,
                               timeout=timeout, verify=verify)
```
That's not the proper final fix but can get the job done until the pydap devs decide the best way for their codebase.

Mikejmnez:
Closed by #310

PR #181, opened 2019-01-03
minor client doc fixes

wholmgren:
A couple of minor doc fixes in `client.rst`:

1. The Timeout section had a couple of urls without ending `'`.
2. Removed a citation that I think is unnecessary at best and offensive at worse.
3. My editor stripped ending whitespace from a handful of lines.


tomkralidis:
Thanks @wholmgren !

Issue #208, opened 2020-04-02
Very slow directory reads for large files / directories (resulting in timeouts)


shreddd:
It looks like pydap is trying to instantiate a `handler` object for every single file that it tries to list when calling get_handler:
https://github.com/pydap/pydap/blob/3f6aa190c59e3bbc6c834377a61579b20275ff69/src/pydap/handlers/lib.py#L72

However the calling function `supported` doesn't actually use this object so it isn't necessary to instantiate a new handler when doing a listing:
https://github.com/pydap/pydap/blob/3f6aa190c59e3bbc6c834377a61579b20275ff69/src/pydap/wsgi/app.py#L151

This slows things to a crawl when trying to list a directory with 100 files - each file is a few GB and takes ~0.5 to 1 sec to open. This ultimately causes timeout issues in the Gunicorn pydap server since it takes over 30 seconds.

A simple fix is to avoid instantiating the object for directory listings (add a parameter to get_handler that bypasses instantiation. 

I have a PR I can submit if this is useful.

PR #209, opened 2020-04-02
Fix to speed up directory listings.

shreddd:
Don't try and create a handler instance for each file in the list (which is expensive).

Fixes Issue #208


captainkirk99:
Is this PR still active? It needs to be updated with the latest version of the master branch...

captainkirk99:
This seems like a good change, if anyone objects, speak now or forever hold your peace...

PR #213, opened 2020-07-29
enable infiity for float attributes

d70-t:
Float attributes should support infinity values.


captainkirk99:
This seems like a good change, and includes tests. Is there any objection to this PR?

captainkirk99:
I have resubmitted these changes as #250 . I will close this PR.

PR #262, opened 2022-03-07
Switch w3.org links to https

shreddd:
http connections to jigsaw.w3.org (and other w3.org URLs are failing) - this should address the network hang caused by this.


Mikejmnez:
Thank you for the PR! It has taken some time for someone from the project to take a look at this but that can happen with open source projects. 

 I took a look at this, and I am not sure if the original problem that prompted this PR/solution is still a problem. I can navigate to the location of `base.html` and do:

```bash
open base.html
```
and it the html file looks fine. Moreover, I take the url http://jigsaw.w3.org/css-validator/ and I see no problem. Perhaps the issue is no longer an issue? (the website opens to a `https` so maybe there is some correction taking place there)/

I am inclined to close this PR without merger. But if you think that this is a problem that persists, or a feature that would enhance to base code, I welcome you to open an issue with a brief and concise (if possible reproducible) explanation of what the undesired behavior it is that this PR would fix.

Thanks again!

PR #272, opened 2022-09-27
Import Mapping from `collections.abc`

rbeucher:
Python versions < 3.10 raise a Deprecation Warning for `from collections import Mapping`

`1: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
`
This means pydap is not working with python 3.10

Solution is to change:

`from collections import Mapping`

to

`from collections.abc import Mapping`

which is the correct way since python 3.3

see: https://stackoverflow.com/a/69727802/13994294


rbeucher:
OK. Looks like my issue was related to the version available on pypi.
Still. I dont think the try/except is needed as this is deprecated since 3.3... unless you want compatibility with python 2??

jgallagher59701 (approved):

jgallagher59701:
@owenlittlejohns I think we can merge this one too. Monday...

src/pydap/tests/test_model.py
```diff
@@ -176,10 +176,7 @@ def test_StructureType_init():
 def test_StructureType_instance():
     """Test that it is a Mapping and DapType."""
     var = StructureType("var")
-    try:
-       from collections.abc import Mapping
-    except ImportError:
-       from collections import Mapping
+    from collections.abc import Mapping
```
owenlittlejohns, line 179: Nit: It would be a little tidier if this import could be just moved up with the main import statements at the top of this module.

owenlittlejohns (commented):
Thanks @rbeucher for this update! I have two super small requests:

* Moving the import statement in `test_model.py` up to the main block of imports at the top of the file.
* There's a similar change that could/should be made in `src/pydap/responses/das.py` for importing `collections.abc.Iterable`.

Thanks again!

owenlittlejohns:
I made the change requested above (updating the import for `collections.abc.Iterable`). I also rebased the branch to be on top of the most recent version of the `master` branch. If the tests all pass, I'll approve and merge this PR. Thanks again @rbeucher for the updates.

PR #275, opened 2022-11-28
Tests fix

jgallagher59701:
Removed python 3.6 from tests.


NiklasPhabian:
I don't understand why the 3.6 is still required even though there is no action for this; might have to do with the branch protection settings?

https://stackoverflow.com/questions/52200096/github-pull-request-waiting-for-status-to-be-reported
https://github.com/orgs/community/discussions/25720

PR #278, opened 2023-03-17
DAP4/DAS Patch

ndp-opendap:
Fixing a typo that was causing the client to attempt to build a DAS in a DAP4 context.

This PR also drops testing for Python 3.6, 3.7, and 3.8. The OS-X tests remain at 3.7 , to be sorted another day.

Some baseline regex had to be updated for newer Python behavior.


jgallagher59701 (approved):
ready to merge

PR #280, opened 2023-04-05
Add workflow to publish pydap to PyPI

owenlittlejohns:
This PR adds a GitHub workflow to enable publication of `pydap` to [PyPI](https://pypi.org/project/Pydap/). This workflow has two potential triggers:
* When a release is published.
* Manual invocation.
Both of these activities should (I think) be limited to maintainers/admins for the repository.

There is one missing piece to this... a new GitHub secret will need to be added to the repository. `PYPI_API_TOKEN`. This is something that needs to be generated by a user account in PyPI, for a user with at least maintainer privileges to the PyPI `pydap` project. @jgallagher59701 has reached out to one of the listed maintainers of the PyPI project to try and become a project maintainer in PyPI.


jgallagher59701 (approved):
Looks good. I think can be merged.

PR #286, opened 2023-07-10
Fix #207 : ImportError: cannot import name 'PasterServerApplication' from 'gunicorn.app.pasterapp'

### LINKED ISSUE #207: ImportError: cannot import name 'PasterServerApplication' from 'gunicorn.app.pasterapp'
Since the last Pypi version (3.2.2) I'm getting this error: 

      File "/home/alberto/anaconda3/bin/pydap", line 10, in <module>
        sys.exit(main())
      File "/home/alberto/anaconda3/lib/python3.7/site-packages/pydap/wsgi/app.py", line 225, in main
        from gunicorn.app.pasterapp import PasterServerApplication
    ImportError: cannot import name 'PasterServerApplication' from 'gunicorn.app.pasterapp' 
    (/home/alberto/anaconda3/lib/python3.7/site-packages/gunicorn/app/pasterapp.py)

I'm running Python 3.7. It seems that the gunicorn 20.0 update broke pydap. Downgrading to gunicorn 19.9 fixes the issue

raphaeljolivet:
I removed the dependency to pasterapp.
This fix works both on newer and older versions of gunicorn (<=, >= to 20.X)

### DIFF
--- a/src/pydap/wsgi/app.py
+++ b/src/pydap/wsgi/app.py
@@ -21,6 +21,7 @@ import mimetypes
 from datetime import datetime
 import shutil
 
+from gunicorn.app.wsgiapp import WSGIApplication
 from jinja2 import Environment, PackageLoader, FileSystemLoader, ChoiceLoader
 from webob import Response
 from webob.dec import wsgify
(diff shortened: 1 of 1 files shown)


src/pydap/wsgi/app.py
```diff
@@ -21,6 +21,7 @@
 from datetime import datetime
 import shutil
 
+from gunicorn.app.wsgiapp import WSGIApplication
```
owenlittlejohns, line 24: Looks like the tests failed for this PR because `gunicorn` is not available for the module to import. I would guess including it in the `install_requires` for `setup.py` would probably mitigate the issue.

src/pydap/wsgi/app.py
```diff
@@ -234,12 +235,31 @@ def init(directory):
         shutil.copy(path, directory)
 
 
+class PyDapApplication(WSGIApplication):
```
owenlittlejohns, line 238: Would it be possible to include a couple of unit tests for this new class?

owenlittlejohns (changes requested):
Thanks for the PR. Looks like there's an import error to fix up and some tests to add, otherwise, it makes sense to me.

Mikejmnez:
I think this is an important PR. You get the same error when you try to use pydap as a server as detailed in the README:
```bash
$ pip install "pydap[server,handlers.netcdf]"
$ pydap --data ./myserver/data/ --port 8001
```
(pydap version 3.4)

@raphaeljolivet If you're too busy and don't mind, I would like to continue working on this PR until merger.

raphaeljolivet:
@Mikejmnez 
Hi, yes, please do.

Mikejmnez:
Awesome, thanks @raphaeljolivet for the much contribution 🚀  ! I will try to modernize `pydap` in the coming weeks and this PR is an important step towards that. I will add you to the list of contributors as you have done must of the work on this one .

Mikejmnez:
closed by #318. Thanks again!

PR #289, opened 2023-09-13
Remove unused `pyOpenSSL` dependency

facutuesca:
`pyOpenSSL` does not seem to be used anywhere in the codebase. This PR removes it from the testing dependencies.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,8 +53,7 @@ tests_extras = [
     'requests-mock',
     'WebTest',
     'flake8',
-    'werkzeug',
-    'pyopenssl']
+    'werkzeug']
 
 
 testing_extras = (functions_extras + cas_extras + server_extras +


Mikejmnez:
Thank you for the heads up! I also tried this myself - somehow macOS depends on it even though there is no explicit import (also all local tests pass on my personal macOS). I will spend some time this week on this.

Mikejmnez:
The trick here was to remove `pyOpenSSL` from the dependencies, but replace it with `cryptography`.  

The appropriate change is being implemented in [pydap/PR#325](https://github.com/pydap/pydap/pull/325), instead of in this branch because too many changes have happened within pydap, and many conflicts must be resolved in this PR branch. Nevertheless, thank you for your contribution @facutuesca!

I will proceed to close this PR.

Issue #292, opened 2024-03-05
`TypeError: cannot pickle '_thread.lock' object` on OSX (and potentially Win32) for python > 3.8


Mikejmnez:
See testing environment  [here](https://github.com/pydap/pydap/files/14488808/env.txt) 

The issue arises when using either of `LocalTestServer` , `LocalTestServerSSL` classes, in combination with `multiprocessing.Process`. These local servers are defined in `src/pydap/server/devel.py` and `src/pydap/server/devel_ssl.py`, and are a "Simple server instance that can be used to test pydap." <ins>HOWEVER</ins>, the `netcdf` class defined in `pydap/api/netcdf4.py` uses the same `LocalTestServer` and therefore this issue is important to address.

## simplest reproducible error:
```python
import numpy as np
from pydap.handlers.lib import BaseHandler
from pydap.model import DatasetType, BaseType
from pydap.server.devel import LocalTestServer

# define the dataset
ExampleDataset = DatasetType("Default")
ExampleDataset["string"] = BaseType("string", np.array(["one", "two"]))
application = BaseHandler(ExampleDataset)
````
The following fails:

```python
server = LocalTestServer(application, as_process=True)
server.start()
```
The TypeError is  :
```python
TypeError: cannot pickle '_thread.lock' object
```
(see full [traceback](https://github.com/pydap/pydap/files/14488964/TracebackError.txt)). This error is somewhat important when working with `netcdf` files because:
```python
ExampleDataset.to_netcdf()
```
yields the same `TypeError`. `ExampleDataset.to_netcdf()` calls `LocalTestServer` with `as_processes=True`.

If `as_process=False`,  `LocalTestServer` uses `threading.Thread` instead, and there is no error at all. 
`as_process=True` is also the default and the `LocalTestServerSSL`. 

## root of the problem
Beginning with `python >= 3.8`, `multiprocessing.Process` uses `spawn` by default to start a process for both windows and OSX
 ([see documentation](https://docs.python.org/3.8/library/multiprocessing.html#contexts-and-start-methods)). I believe this is the reason why the tests failed for python version 3.8+ on PR #283  for the macOS testing environment. 

## potential patchworks
a) 
This is by no means any long term solution, but setting "fork" before starting the process (`LocalTestServer` with `as_process=True`) 
```python
from multiprocessing import set_start_method
set_start_method("fork", force=True)
```
[...]

Issue #256, opened 2022-01-31
pydap tests fail on macos for python >= 3.8


captainkirk99:
```
Traceback (most recent call last):
119
  File "<string>", line 1, in <module>
120
  File "/Users/runner/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/multiprocessing/spawn.py", line 116, in spawn_main
121
    exitcode = _main(fd, parent_sentinel)
122
  File "/Users/runner/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/multiprocessing/spawn.py", line 126, in _main
123
    self = reduction.pickle.load(from_parent)
124
  File "/Users/runner/work/pydap/pydap/src/pydap/model.py", line 238, in __getattr__
125
    return self.attributes[attr]
126
  File "/Users/runner/work/pydap/pydap/src/pydap/model.py", line 238, in __getattr__
127
    return self.attributes[attr]
128
  File "/Users/runner/work/pydap/pydap/src/pydap/model.py", line 238, in __getattr__
129
    return self.attributes[attr]
130
  [Previous line repeated 993 more times]
131
RecursionError: maximum recursion depth exceeded
132
src/pydap/server/devel_ssl.py F                                          [  3%]
```

captainkirk99:
This happens for python 3.8 and greater. Python 3.7 succeeds in all tests on macos.

headtr1ck:
Same happens on windows with python 3.9

aulemahal:
Same happening on Linux, pydap 3.3.0, python 3.8.13

I am using pydap through xarray, which uses dask to perform parallel computation. The problem arises when using dask in a multi-process configuration and it does not arise when using the multi-threads version.
`Dask`, alike `multiprocessing`, `pickle` to communicate objects between different workers (processes).

Here is an example.
```python3
with dask.config.set(num_workers=2, scheduler='processes'):
    public_url = 'http://psl.noaa.gov/thredds/dodsC/Datasets/ncep.reanalysis/Monthlies/surface/air.mon.ltm.nc'
    ds = xr.open_dataset(
        public_url,
        engine='pydap', user_charset='utf8',
        chunks={'lon': 12}
    )
    mean = ds.air.mean(['lon', 'lat']).load()
```
which fails with:
```python3
---------------------------------------------------------------------------
RecursionError                            Traceback (most recent call last)
[...]

PR #293, opened 2024-03-05
Allow newer python versions to test on MacOS mimicking Ubuntu workflows

Mikejmnez:
This PR 
- [x] Provides a temporal patch to the pickling issues described in #292. A better long term solution needs to address this. 
- [x] Allows `pydap` to work with newer python version in both macOS and Windows env.
- [x] Updates the MacOS Python versions that the tests run on to align with the Ubuntu workflow. Thus addresses #256. 
- [x] All tests pass on my local testing environment with `python=3.10`

There was an earlier effort to update the python versions for testing workflow on #283. This PR essentially allows that to happen. However, this PR does not address the issue raised in #292 with `multiprocessing.Process and netcdf`. But it allows for updated testing workflow by skipping tests that may fail on MacOS (there is no issue on Ubuntu / linux). A separate PR should address #292.

Suggestions are welcome. This can also be a draft PR, but I do think that there is merit in updating the testing (python) environment, while also identifying the issues with MacOS and Windows

### DIFF
--- a/src/pydap/server/devel.py
+++ b/src/pydap/server/devel.py
@@ -47,40 +47,40 @@ class LocalTestServer(object):
     Relies on threading and is usually slow (it has to
     start and shutdown which typically takes ~2 sec).
 
-    # Usage:
-    # >>> import numpy as np
-    # >>> from pydap.handlers.lib import BaseHandler
-    # >>> from pydap.model import DatasetType, BaseType
-    # >>> DefaultDataset = DatasetType("Default")
-    # >>> DefaultDataset["byte"] = BaseType("byte", np.arange(5, dtype="B"))
-    # >>> DefaultDataset["string"] = BaseType("string", np.array(["one", "two"]))
-    # >>> DefaultDataset["short"] = BaseType("short", np.array(1, dtype="h"))
-    # >>> DefaultDataset
-    # <DatasetType with children 'byte', 'string', 'short'>
-    # >>> application = BaseHandler(DefaultDataset)
-    # >>> from pydap.client import open_url
+    Usage:
+    >>> import numpy as np
+    >>> from pydap.handlers.lib import BaseHandler
+    >>> from pydap.model import DatasetType, BaseType
+    >>> DefaultDataset = DatasetType("Default")
+    >>> DefaultDataset["byte"] = BaseType("byte", np.arange(5, dtype="B"))
+    >>> DefaultDataset["string"] = BaseType("string", np.array(["one", "two"]))
+    >>> DefaultDataset["short"] = BaseType("short", np.array(1, dtype="h"))
+    >>> DefaultDataset
+    <DatasetType with children 'byte', 'string', 'short'>
+    >>> application = BaseHandler(DefaultDataset)
+    >>> from pydap.client import open_url
 
-    # As an instance:
-    # >>> with LocalTestServer(application) as server:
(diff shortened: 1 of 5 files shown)


Mikejmnez:
This PR should also help address for now the long unresolved downstream issue in: [pydata/xarray/#4348](https://github.com/pydata/xarray/issues/4348)

src/pydap/server/devel.py
```diff
@@ -46,40 +47,40 @@ class LocalTestServer(object):
...
-    Or by managing connection and deconnection:
-    >>> server = LocalTestServer(application)
-    >>> server.start()
-    >>> dataset = open_url("http://localhost:%s" % server.port)
-    >>> dataset
-    <DatasetType with children 'byte', 'string', 'short'>
-    >>> print(dataset['byte'].data[:])
-    [0 1 2 3 4]
-    >>> server.shutdown()
+    # Usage:
```
jgallagher59701, line 50: Why comment out the usage info? If it is wrong, can you say why?

src/pydap/server/devel_ssl.py
```diff
@@ -44,36 +45,36 @@ class LocalTestServerSSL(LocalTestServer):
...
-    >>> import numpy as np
-    >>> from pydap.handlers.lib import BaseHandler
-    >>> from pydap.model import DatasetType, BaseType
-    >>> DefaultDataset = DatasetType("Default")
-    >>> DefaultDataset["byte"] = BaseType("byte", np.arange(5, dtype="B"))
-    >>> DefaultDataset["string"] = BaseType("string", np.array(["one", "two"]))
-    >>> DefaultDataset["short"] = BaseType("short", np.array(1, dtype="h"))
-    >>> DefaultDataset
-    <DatasetType with children 'byte', 'string', 'short'>
+    # Usage:
```
jgallagher59701, line 48: ditto

jgallagher59701 (approved):
Look at my two comments and let me know. Aside from that, this looks ready to merge. 

Lets also talk about a long term solution. My Python knowledge is limited and I wonder if moving from  WSGI to Flask might be a good move overall since that might reduce the amount of lower level configuration we have to do.

src/pydap/server/devel.py
Mikejmnez: I commented this because it was producing some errors in my local testing environment. That was on an earlier version of the PR. I completely forgot about these comments.

src/pydap/server/devel_ssl.py
Mikejmnez: I commented this because it was producing some errors in my local testing environment. That was on an earlier version of the PR. I completely forgot about these comments.

Mikejmnez:
> Lets also talk about a long term solution. My Python knowledge is limited and I wonder if moving from WSGI to Flask might be a good move overall since that might reduce the amount of lower level configuration we have to do.

Yeah. That makes sense to me. That should take place on a separate issue/PR.

src/pydap/server/devel.py
Mikejmnez: uncommented. tests passing. I will mark this as resovled

src/pydap/server/devel_ssl.py
Mikejmnez: uncommented. test continue to pass. I will mark this as resolved

Issue #303, opened 2024-04-04
remove six dependency


Mikejmnez:
### Description

[six](https://pypi.org/project/six/) provides compatibility with python 2.7 and 3.x. However pydap only support now python  >= 3.7. We should remove this dependency, refactor and replace it with an updated dependency for 3.x versions.

The following are the places where `six` is imported and all the ways it is being used:
```
(base) $ grep -r "six"
./cas/esgf.py:from six.moves.urllib.parse import quote_plus
./cas/get_cookies.py:from six.moves.urllib.parse import urlsplit, urlunsplit
./parsers/__init__.py:from six.moves.urllib.parse import unquote
./parsers/__init__.py:from six.moves import map
./parsers/das.py:from six.moves import reduce
./lib.py:from six.moves.urllib.parse import quote as quote_
./lib.py:from six.moves import reduce, zip_longest
./lib.py:from six import binary_type, MAXSIZE, string_types
./client.py:from six.moves.urllib.parse import urlsplit, urlunsplit
./wsgi/ssf.py:from six.moves import reduce, map
./wsgi/ssf.py:from six import string_types
./wsgi/app.py:from six.moves.urllib.parse import unquote
./wsgi/app.py:from six import string_types
./net.py:from six.moves.urllib.parse import urlsplit, urlunsplit
./responses/version.py:from six import text_type
./responses/error.py:from six import StringIO, text_type
./responses/dods.py:from six import string_types
./responses/ascii.py:from six.moves import zip
./responses/html/__init__.py:from six.moves.urllib.parse import unquote
./responses/das.py:from six import string_types, integer_types
./responses/das.py:from six.moves import map
./responses/dds.py:from six.moves import map, zip
./tests/test_wsgi_ssf.py:from six import next
./tests/test_lib.py:from six import MAXSIZE
./tests/test_handlers_netcdf.py:from six.moves import zip
./tests/test_handlers_lib.py:from six import text_type
./model.py:from six.moves import reduce, map
./model.py:from six import string_types
./handlers/dap.py:import six.moves.urllib.parse
./handlers/dap.py:from six import text_type, string_types, BytesIO
./handlers/dap.py:        scheme, netloc, path, query, fragment = six.moves.urllib.parse.urlsplit(self.url)
./handlers/dap.py:        self.base_url = six.moves.urllib.parse.urlunsplit(arg)
[...]

PR #304, opened 2024-04-05
Iss303

Mikejmnez:
The following Pull Request:

- [x] Closes issue #303 
- [x] All tests pass on a local environment. No need to add extra tests.

I followed [this as guide](https://readthedocs.org/projects/six/downloads/pdf/stable/) to understand how to remove all explicit `six` dependency.

The package `six` provides compatibility with python 2.7 and python 3.x versions. And so removing `six` is a refactor since pydap is no longer compatible with python versions <3.7


jgallagher59701 (approved):
This looks ready to merge.
Thanks!

Issue #305, opened 2024-04-05
add project.toml file


Mikejmnez:
### Description

Nowadays python strongly recommends to have a project.toml file and in particular the `[buid-system]` configuration. Also, regarding setup.py, a lot of the functionality it used to provide is depracated. For more on this read [HERE](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/)

Migrating much of the content of setup.py to project.toml will involve:

- [ ] correctly defining entry points.
- [ ] correctly defining implicit namespaces. There are a lot of warnings already about this. See #277 .
- [ ] correctly building the wheels and tar files.

Mikejmnez:
closed by #307

PR #307, opened 2024-04-06
migrating setup.py to pyproject.toml

### LINKED ISSUE #277: namespace_packages deprecated
namespace_packages as defined in 
https://github.com/pydap/pydap/blob/35077495efc253116bf0d08f99b56eec82135bad/src/pydap/tests/__init__.py#L1

are deprecated. See also https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages and matplotlib/matplotlib#25244

Mikejmnez:
The following Pull Request:

- [x] Closes issue #305 
- [x] Also closes #277. In closing this issue, I had to make some modificiations to allow for namespaces and package recovery (which is now implicit). That meant:
- [x] Removed `__init__.py` file within `src/pydap/handlers/`.  For more information about this about [namespaces](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/) and [package recovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#custom-discovery).
- [x] A new feature is added. It is now easy to print pydap's version (which is inferred from commits).
- [x] A test is included to read the version number.

As described in #305, much of the functionality of setup.py has been deprecated and at the very least there should always be an pyproject.toml file. I went further and migrated much of setup.py to pyproject.toml. 

```python
import pydap
pydap.__version__
>>> version_number
```
When a new version of pydap is released (say 3.4.2),
```bash
% python-c "import pydap; print(pydap.__version__)"
>>> 3.4.2
```
The versioning is generated during installation. When pydap is installed  --editable mode
```python
% pip install -e .
% python -c "import pydap; print(pydap.__version__)"
>>> 0.1.dev680+g61f0c70.d20240321
```
or so, becoming clear that the developer is using an editable version. This is common practice.


Mikejmnez:
This PR is ready to be merged.

pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+
+
+[tool.setuptools]
+package-dir = {"" = "src"}
+
+[tool.setuptools.packages.find]
+where = ["src"]
+
+[tool.setuptools_scm]
+version_file = "src/pydap/_version.py"
```
jgallagher59701, line 98: Add a new line at the end of this file

src/pydap/__init__.py
```diff
@@ -1,6 +1,7 @@
...
-'''
-__import__('pkg_resources').declare_namespace(__name__)
+from importlib.metadata import version, PackageNotFoundError
 
-__version__ = '3.4.1'
+try:
+    __version__ = version("pydap")
+except PackageNotFoundError:
+    # package is not installed
+    pass
```
jgallagher59701, line 7: Same this here

jgallagher59701 (approved):
Two minor edits - thanks for the namespace fixes and packing support.

pyproject.toml
Mikejmnez: done

src/pydap/__init__.py
Mikejmnez: done

PR #310, opened 2024-04-12
Allow escaping of '[' and ']' characters when opening remote datasets with dap4 protocol

### LINKED ISSUE #306: URL escaping: Does PyDAP escape characters in URLs?
### Description

Does PyDAP escape the characters in the URL? It might be that the Requests package takes care of that for you (that’s great if true). We have this issue with Sam’s PyDMR test system where raw URLs are used and things like `dap4_ce=/time[0]` break tomcat-based web servers (because characters such as ‘[’ are not allowed in a URL).

### Minimal example that reproduces the bug, or a new desired feature

This URL will fail if pasted into a browser: `https://opendap.earthdata.nasa.gov/collections/C1276812863-GES_DISC/granules/M2T1NXSLV.5.12.4%3AMERRA2_100.tavg1_2d_slv_Nx.19800101.nc4.dap?dap4.ce=/time[0]`
because the `...time[0]` at the end of the URL contains `[` and `]` which are (formally) not allowed to be in a URL. 

NB: That URL will need to be modified for use with PyDAP to something like: `https://opendap.earthdata.nasa.gov/collections/C1276812863-GES_DISC/granules/M2T1NXSLV.5.12.4%3AMERRA2_100.tavg1_2d_slv_Nx.19800101.nc4`
and then use DAP4 to ask for the first element of the variable `/time`

If the URL is fed into a browser with the `[` etc., _escaped_, then it will work fine. That would look like:
`https://opendap.earthdata.nasa.gov/collections/C1276812863-GES_DISC/granules/M2T1NXSLV.5.12.4%3AMERRA2_100.tavg1_2d_slv_Nx.19800101.nc4.dap?dap4.ce=/time%5B0%5D`
(0x5B is hex for '[', ..., and that's how URL escaping works.

PyDAP will need to escape characters in the Constraint Expression (CE)

Mikejmnez:
The following Pull Request:

- [x] Closes #306. It may also close PR #180 
- [x] Tests were included and pass locally in test environment.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -134,6 +134,8 @@ class DAPHandler(pydap.handlers.lib.BaseHandler):
                 self.fragment,
             )
         )
+        dmr_url = dmr_url.replace("[", "%5B").replace("]", "%5D")
+
         r = pydap.net.GET(
             dmr_url,
             self.application,
@@ -438,11 +440,14 @@ class BaseProxyDap4(BaseProxyDap2):
 
     def __getitem__(self, index):
         # build download url
-        index = combine_slices(self.slice, fix_slice(index, self.shape))
-        scheme, netloc, path, params, query, fragment = urlparse(self.baseurl)
-        ce = "dap4.ce=" + quote(self.id) + hyperslab(index) + query
-        url = urlunparse((scheme, netloc, path + ".dap", "", ce, fragment)).rstrip("&")
 
+        index = combine_slices(self.slice, fix_slice(index, self.shape))
+        scheme, netloc, path, _, query, fragment = urlparse(self.baseurl)
+        # ce = "dap4.ce=" + quote(self.id) + hyperslab(index) + query
+        # else:
+        ce = "dap4.ce=" + quote(self.id) + hyperslab(index)
+        url = urlunparse((scheme, netloc, path + ".dap", "", ce, fragment)).rstrip("&")
+        # I need to encode `[` and `]`
         # download and unpack data
         logger.info("Fetching URL: %s" % url)
 
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@ pydap
 [![black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![license](https://img.shields.io/github/license/mashape/apistatus.svg)](https://github.com/pydap/pydap)
 [![pre-commit](https://results.pre-commit.ci/badge/github/pydap/pydap/master.svg)](https://results.pre-commit.ci/latest/github/pydap/pydap/master)
-[![Join the chat at https://gitter.im/pydap/pydap](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/pydap/pydap)
 
 
 
--- a/src/pydap/tests/test_client.py
+++ b/src/pydap/tests/test_client.py
@@ -31,11 +31,24 @@ def test_open_url(sequence_app):
 
 
 def test_open_url_dap4():
-    url = "http://test.opendap.org/opendap/hyrax/data/nc/test.nc"
-    constrain = "dap4.ce=/s33[0][0]"
-    data_original = open_url(url)
-    data_dap4 = open_url(url + "?" + constrain, protocol="dap4")
+    base_url = "http://test.opendap.org/opendap/hyrax/data/nc/test.nc"
+    data_original = open_url(base_url)
+
+    # test single data point
+    constrain1 = "dap4.ce=/s33[0][0]"
+    data_dap4 = open_url(base_url + "?" + constrain1, protocol="dap4")
     assert data_dap4["s33"][:].data == data_original["s33"][0, 0].data
+
+    # subset of vars
+    var1 = "/s33[0:][0:1:2];"
+    var2 = "/br34[0:1:1][0:1:2][0:1:3];"
+    var3 = "/s113[0:1:0][0:1:0][0:1:2]"
+    Vars = [var1, var2, var3]
+
+    url = base_url + "?dap4.ce=" + var1 + var2 + var3
+    dataset = open_url(url, protocol="dap4")
+    # check [vars1, vars2, vars3] only in dataset
+    assert len(dataset.keys()) == len(Vars)
 
 
 @pytest.mark.client
--- a/src/pydap/tests/test_client.py
+++ b/src/pydap/tests/test_client.py
@@ -30,6 +30,14 @@ def test_open_url(sequence_app):
     assert list(dataset.keys()) == ["cast"]
 
 
+def test_open_url_dap4():
+    url = "http://test.opendap.org/opendap/hyrax/data/nc/test.nc"
+    constrain = "dap4.ce=/s33[0][0]"
+    data_original = open_url(url)
+    data_dap4 = open_url(url + "?" + constrain, protocol="dap4")
+    assert data_dap4["s33"][:].data == data_original["s33"][0, 0].data
+
+
 @pytest.mark.client
 def test_open_dods():
     """Open a file downloaded from the test server with the DAS."""


jgallagher59701 (changes requested):
Also, this needs to be done for DAP2 as well as DAP4.

Thanks.

src/pydap/handlers/dap.py
```diff
@@ -134,6 +134,8 @@ def dataset_from_dap4(self):
                 self.fragment,
             )
         )
+        dmr_url = dmr_url.replace("[", "%5B").replace("]", "%5D")
```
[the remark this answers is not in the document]
Mikejmnez: python's `request.urlparse` and `request.urlunparse` can split an url into 6 different component just like `urllib` in that Gemini example. That already was taking place before the new encoding feature of `[` and `]`. 

Here is the workflow of `open_url`:
1) (<ins>**Unchanged**</ins>) Pass the `url`  to the `DAPHandler` class, where it gets parsed and unparsed to determine the protocol (dap2 vs dap4). Relevant code here:

> https://github.com/pydap/pydap/blob/15afa3d0f2731965fcdfc89909bde3e46872452e/src/pydap/handlers/dap.py#L77-L98

The relevant `scheme`, `netloc`, `path`, `query`, etc of the `url` are assigned to `self.` so they become callable within the class. Note that the original `url` is now assigned to `base_url` and also becomes callable.

2) (<ins>**Changed**</ins>)`make_dataset` (line 96 above) creates a `DatasetType` with dummy variables. This `DatasetType` must have the right number of variables according to the query/constraint in the url. The important piece here is the `dmr_url`. Here is where `pydap` needs to escape the `[` `]` to generate the `dmr_url` accurately. For that, this PR takes the dmr_url and escapes '[' and ''].

For example, the test url generated from the DAP Access Form
```python
url = "http://test.opendap.org/opendap/hyrax/data/nc/test.nc?dap4.ce=/s33[0:][0:1:2];/s113[0:1:0][0:1:0][0:1:2]"
```
now produces the following escaped `drm_url`:

```python
"http://test.opendap.org/opendap/hyrax/data/nc/test.nc.dmr?dap4.ce=/s33%5B0:%5D%5B0:1:2%5D;/s113%5B0:1:0%5D%5B0:1:0%5D%5B0:1:2%5D"
```
If you  copy-paste into a browser, it takes you to the DAP Access Form which contains the right metadata.

3) (<ins>**Changed**</ins>) `add_proxies()`  (line 98 above) adds a proxy to the DatasetType  per variable excluding lat, lon, for the remote data set url.
<int>The change I did was to not append the variable `query`to the `ce` string</ins>. You can see that in the PR. The reason behind this was because of "constraint expreesion parse errors". 
Here is the original relevant code, which takes place within a loop per each variable in the DatasetType:
```python
# this is within a loop per each variable named self.id
scheme, netloc, path, params, query, fragment = urlparse(self.baseurl)
ce = "dap4.ce=" + quote(self.id) + hyperslab(index) + query
```
When there is no constraint, `query=''` and it adds nothing to ce. 
When there is a constraint,  `query` contains the constraint of the original file (`base_url`). This messes up the constraint expression for each `self.id` variable within the loop.

If query is NOT removed, I get the following error:

```python
HTTPError: 400 400
<?xml version="1.0" encoding="UTF-8"?>
<dap4:Error xmlns:dap4="http://xml.opendap.org/ns/DAP/4.0#" httpcode="400">
  <dap4:Message>libdap error transmitting DAP4Data: Constraint expression parse error: syntax error, unexpected word, expecting end of file</dap4:Message>
  <dap4:Context [/](http://localhost:8888/)>
  <dap4:OtherInformation [/](http://localhost:8888/)>
<[/dap4](http://localhost:8888/dap4):Error>
```

And so query string of the original `base_url` should not be added to `ce`.

Mikejmnez:
The tests added for this PR use the following url: 

```python
http://test.opendap.org/opendap/hyrax/data/nc/test.nc?dap4.ce=/s33[0:][0:1:2];/br34[0:1:1][0:1:2][0:1:3];/s113[0:1:0][0:1:0][0:1:2]
```

src/pydap/handlers/dap.py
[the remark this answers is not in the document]
Mikejmnez: I will look at other potential characters that should be appropriately escaped, and include proper example of desired behavior that this PR should incorporate.

src/pydap/handlers/dap.py
```diff
@@ -439,8 +441,8 @@ def __repr__(self):
     def __getitem__(self, index):
         # build download url
         index = combine_slices(self.slice, fix_slice(index, self.shape))
-        scheme, netloc, path, params, query, fragment = urlparse(self.baseurl)
-        ce = "dap4.ce=" + quote(self.id) + hyperslab(index) + query
+        scheme, netloc, path, _, query, fragment = urlparse(self.baseurl)
+        ce = "dap4.ce=" + quote(self.id) + hyperslab(index)
```
Mikejmnez, line 445: I removed this `query` which comes from the original dataset, because it is unnecessary for two reasons:
1. `query=''` when there is non constraint in the original `url`. 
2. `query` contains global (all variables) information about constraints. This snipped occurs within a loop per each variable id. The constraint for each variable is already contained within quote(self.id) and hyperslab(index). Leaving `query` leads to the following 
```python
HTTPError: 400 400
<?xml version="1.0" encoding="UTF-8"?>
<dap4:Error xmlns:dap4="http://xml.opendap.org/ns/DAP/4.0#" httpcode="400">
  <dap4:Message>libdap error transmitting DAP4Data: Constraint expression parse error: syntax error, unexpected word, expecting end of file</dap4:Message>
  <dap4:Context [/](http://localhost:8888/)>
  <dap4:OtherInformation [/](http://localhost:8888/)>
<[/dap4](http://localhost:8888/dap4):Error>
```

Mikejmnez:
Draft mode - some changes still necessary

Mikejmnez:
### Description of state of this (active) PR

After commit [ebefa0f](https://github.com/pydap/pydap/pull/310/commits/ebefa0f72447dc0bf1a665efe927c1fd42a60466), this now works (before it didn't):

* **DAP2**: Sequences and constraint expressions:

```python
$ url = "http://test.opendap.org/opendap/data/ff/1998.6.avhrr.dat"
$ ds = open_url(url)
$ ds['GSO_AVHRR.hh']
>>> <BaseType with data SequenceProxy('http://test.opendap.org/opendap/data/ff/1998.6.avhrr.dat', <BaseType with data BaseProxy('http://test.opendap.org/opendap/data/ff/1998.6.avhrr.dat', 'GSO_AVHRR.hh', dtype('>i4'), (), ())>, [], (slice(None, None, None),))>
```
In particular the following didn't work (see issue #313)
```python
$ dsc = open_url(url+'?GSO_AVHRR.year,GSO_AVHRR.day_num,GSO_AVHRR.hh&GSO_AVHRR.hh>12')
$ np.array([val for val in ds['GSO_AVHRR']['hh']])
>>> array([21, 22, 22, 21, 21, 22, 22, 21, 21, 21, 22, 22, 21, 21, 22, 22, 21,
       21, 22, 21, 21, 21, 22, 21, 21, 21, 22, 22], dtype=int32)
```

* **DAP4**: Groups implements fully qualifying names correctly (nested groups+sequence/structures), and constraint expression for Groups

The following works properly now (it didn't) -- NOTE: there is no need to specify protocol (dap4) anymore!!
```python
$ url1 = 'http://test.opendap.org:8080/opendap/dmrpp_test_files/ATL03_20181228015957_13810110_003_01.2var.h5.dmrpp?dap4.ce=/gt1r/heights/lat_ph[0:1:100];/gt1r/heights/delta_time[0:1:100]'
$ ds = open_url(url1)
$ ds
>>> <DatasetType with children 'gt1r'>
```
where `gt1r` is a Group (see past behavior explained in issue #314). There is a new `GroupType` that acts as container.

In the example before, you can inspect the arrays directly from root with a path-like key:
```python
$ ds['/gt1r/heights/delta_time'].shape
>>> (101,)
```
And to see that the constraint expression works:
```python
$ ds['/gt1r/heights/delta_time'][:10]
>>> <BaseType with data array([31197611.26348985, 31197611.26408982, 31197611.26478983,
       31197611.26608984, 31197611.26758983, 31197611.26888984,
       31197611.27188982, 31197611.27728982, 31197611.27748981,
       31197611.27778983])>
```

src/pydap/handlers/dap.py
jgallagher59701: Ok. I don't enough about Python and OOP to make much of comment either way.

src/pydap/parsers/dmr.py
```diff
@@ -40,7 +41,7 @@ def dap4_to_numpy_typemap(type_string):
     return np.dtype(dtype_str)
 
 
-def get_variables(node, prefix=""):
+def get_variables(node, prefix="") -> dict:
```
jgallagher59701, line 44: Do you plan to add type qualifiers to all the methods? Just curious since I've used them and found them useful, but I can see that adding them could be a _huge_ effort. If you do plan on that, my suggestion would be to make it a separate PR.

jgallagher59701 (commented):
Look like you're making good progress.

jgallagher59701:
Ah, I see now, you probably want to merge master before going much further.

src/pydap/parsers/dmr.py
Mikejmnez: Yes, I think it would be very useful to add the type qualifiers to the methods on a separate PR as you mention.

Mikejmnez:
Thanks for your comments @jgallagher59701 . I agree with your overall assessment: stop at DAP4 + escaping of `[` `]` characters. When I started to work on this branch / PR (to correctly escaping characters), I didn't know all the changes that were needed first in order to allow escaping of all possible expressions / characters. 

I need to correct some tests, and then merge master with this branch. Afterwards, it should be ready for merging.

jgallagher59701:
Good plan.

We can make tickets for the remaining issues - completing the escaping, adding more conformance to DAP4 as those issues are revealed - and push work on those after this code is merged.

Mikejmnez:
@jgallagher59701 well finally this is ready for merger. I wrote an entire new pydap response utility that generates a DMR text so that pydap, oin theory, can server Group data. I added some basic testing for it. It, however, only generates a DMR text for arrays and/or Groups data.  That is, sequences and structures need to be added, but this will be on a separate PR. This one is already too long and old.

This PR also pins numpy to version < 2.0 (see #321). I also need to work on this (not a very quick fix) on a separate PR.

Finally, I will work on a separate PR for the remaining issue of streaming Sequence data with DAP4 (the remaining aspect missing from the original objective of this PR).

Mikejmnez:
sorry - that last commit was about reverting a change to a string representation in a test. But because the commit returns the string representation to what it used to be and so it doesn't even show anymore.
### before
```python
"string_url?"+ "query" # at some point I tested escaping the query, which required me to split the string - but then decided not so escape the query component but forgot to return the string representation to what it used to be (see below)
```
### Now (but also what it used to be)
reverted to original
```
"string url?query"
```

src/pydap/handlers/dap.py
```diff
@@ -105,13 +107,22 @@ def determine_protocol(self, protocol):
...
+            self.sheme = "http"
+            return "dap4"
         else:
             extension = self.path.split(".")[-1]
             if extension in ["dmr", "dap"]:
                 return "dap4"
             elif extension in ["dds", "dods"]:
                 return "dap2"
             else:
+                _warnings.warn(
```
Mikejmnez, line 120: This is in case protocol is None. Then pydap picks DAP2 (because at this point more of DAP2 is functional within pydap. But I figure that this warning should inform people about what they are using. Later on, we should make DAP4 the default.

src/pydap/handlers/dap.py
```diff
@@ -228,6 +244,7 @@ def add_dap2_proxies(self):
             var.data = SequenceProxy(
                 self.base_url,
                 template,
+                selection=self.selection,
```
Mikejmnez, line 247: This was completely missing even for DAP2. Now it is being passed and constraints expression for sequences work as expected (DAP2).

src/pydap/handlers/dap.py
```diff
@@ -439,8 +456,8 @@ def __repr__(self):
     def __getitem__(self, index):
         # build download url
         index = combine_slices(self.slice, fix_slice(index, self.shape))
-        scheme, netloc, path, params, query, fragment = urlparse(self.baseurl)
-        ce = "dap4.ce=" + quote(self.id) + hyperslab(index) + query
+        scheme, netloc, path, _, query, fragment = urlparse(self.baseurl)
+        ce = "dap4.ce=" + self.id + hyperslab(index)
```
Mikejmnez, line 460: `query` does not need to be passed  - variable names are already incorporated within `self.id`

Mikejmnez: also the `_` implies the third element of a tuple is released from memory (not assigned to any variable)

src/pydap/handlers/dap.py
```diff
@@ -813,8 +830,9 @@ def unpack_dap4_data(xdr_stream, dataset):
     buffer = stream2bytearray(xdr_stream)
 
     start = 0
-    for variable_name in dataset:
-        variable = dataset[variable_name]
+    for variable in walk(dataset, pydap.model.BaseType):
```
Mikejmnez, line 833: extracts only variables of BaseType (and not say Groups) which are GroupType

src/pydap/lib.py
```diff
@@ -288,7 +309,7 @@ def fix_shorthand(projection, dataset):
         if len(var) == 1 and var[0][0] not in list(dataset.keys()):
             token, slice_ = var.pop(0)
             for child in walk(dataset):
-                if token == child.name:
+                if token == unquote(child.name):
```
Mikejmnez, line 312: I found this necessary. Otherwise when there was a projection of a variable name, say "foo[", child.name was escaped but "foo[" in `token` (taken form the query string) was not being recognized to be part of the original dataset. An alternative was to escape (or quote) token in the comparison. That also worked. The variable name in `token` is later escaped.

src/pydap/model.py
```diff
@@ -258,6 +265,10 @@ def __init__(
     def __repr__(self):
         return "<%s with data %s>" % (type(self).__name__, repr(self.data))
 
+    @property
```
Mikejmnez, line 268: This is useful when the variable names is a path-like (i.e. groups, sequences, etc)

src/pydap/model.py
```diff
@@ -503,15 +514,81 @@ class DatasetType(StructureType):
         >>> dataset["B"] = BaseType("B")
         >>> dataset["B"].id
         'B'
-
     """
 
     def __setitem__(self, key, item):
-        StructureType.__setitem__(self, key, item)
+        # StructureType.__setitem__(self, key, item)
```
Mikejmnez, line 520: This allows to assign/define a variable within a Dataset with a name that is like
```
Dataset['/Group1/Group2/var_name'] = BaseType('var_name')
```

src/pydap/model.py
```diff
@@ -503,15 +514,81 @@ class DatasetType(StructureType):
...
+                del self[key]
+
+            self._dict[key] = item
+            # By default added keys are visible:
+            self._visible_keys.append(key)
 
         # The dataset name does not go into the children ids.
         item.id = item.name
 
+    def _getitem_string(self, key):
```
Mikejmnez, line 556: This allows to recover a variable from a name that is path-like. For example:
```python
$ dataset['Group1/Group2/var_name']
>>> BaseType('var_name')
```

src/pydap/parsers/__init__.py
```diff
@@ -110,16 +108,21 @@ def parse_ce(query_string):
     Returns a tuple with the projection and the selection.
 
     """
-    tokens = [token for token in unquote(query_string).split("&") if token]
-    if not tokens:
+    if protocol == "dap2":
```
Mikejmnez, line 111: This parsing of projections and selections only work for DAP2, since there is an ampersan there, but this part is only applicable for Sequences and Structures. A different PR will fix this to allow this behavior on DAP4 for Seqs and Structs

src/pydap/parsers/dmr.py
```diff
@@ -80,12 +81,15 @@ def get_dtype(element):
...
 
-def get_attributes(element):
-    attributes = {}
+def get_attributes(element, attributes={}):
     attribute_elements = element.findall("Attribute")
+    numType = [item for item in dmr_atomic_types if item not in ["Byte", "Char"]]
     for attribute_element in attribute_elements:
         name = attribute_element.get("name")
         value = attribute_element.find("Value").text
+        _type = attribute_element.get("type")
```
Mikejmnez, line 90: This bit comes in handy for the cases where Missing value or Fill_value have a Value that is of string type. To follow CF conventions, the type of Missing_value and Fill_value must of a numeric type.

src/pydap/parsers/dmr.py
```diff
@@ -169,35 +177,80 @@ def dmr_to_dataset(dmr):
         for dim in variable["dims"]:
             variable["shape"] += (variables[dim]["size"],)
 
-    # Convert the ordered dictionary to dataset
-    dataset_name = dom_et.attrib["name"]
-    dataset = pydap.model.DatasetType(dataset_name)
     for name, variable in variables.items():
-        data = DummyData(dtype=variable["dtype"], shape=variable["shape"])
+        var_name = variable["name"]
```
Mikejmnez, line 181: This makes it possible to separate the variable name from a path-like variable name like `Group1/Group2/var_name`. path is later passed as an attribute to the specific variable, which is then used to correctly assign Group behavior in the dataset

src/pydap/parsers/dmr.py
```diff
@@ -169,35 +177,80 @@ def dmr_to_dataset(dmr):
...
-            var[name] = array
+            var = pydap.model.GridType(name=var_name)
+            var[var_name] = array
             for dim in variable["dims"]:
                 var[dim] = copy.copy(dataset[dim])
         else:
             var = array
         var.attributes = variable["attributes"]
-        dataset[var.name] = var
+        if "parent" in variable.keys() and variable["parent"] in [
```
Mikejmnez, line 208: This works in the case variable name is oath like, but the variable belongs to a sequence or structure. For example:
```python
Group1/Group2/Seq_name/Var_name
```
Then the variable name is defined as `Seq_name.Var_name`, and path='Group1/Group2'. Similarly for a structure

src/pydap/responses/dmr.py
```diff
@@ -0,0 +1,113 @@
+"""The DMR response.
```
Mikejmnez, line 1: Builds a xml representation of the data, so that pydap can serve a dataset containing arrays and groups. Does not yet work with Sequences and Structures. That will be another PR

src/pydap/tests/data/dmrs/SimpleGroup.dmr
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
```
Mikejmnez, line 1: Sample xml document. pydap.responses.dmr is able to generate this file from an array created interactively. I added the comparison (assertEqual) as a test

src/pydap/tests/datasets.py
```diff
@@ -217,6 +224,23 @@
     "y", np.arange(2), axis="Y", units="degrees_north"
 )
 
+SimpleGroup = DatasetType("example dataset", description="A simple group for testing.")
```
Mikejmnez, line 227: This Simple Group is the one that generates the dmr response (xml) above.

src/pydap/tests/test_handlers_dap.py
```diff
@@ -258,22 +260,8 @@ def test_custom_timeout_SequenceProxy(self):
 
     def test_custom_timeout_BaseProxyDap4(self):
```
Mikejmnez, line 261: No longer needed to monkeypath a DAT4 dataset from a DAP2 one. :)

jgallagher59701 (approved):
This is ready to be merged. Thanks!

PR #322, opened 2024-07-11
Np2

### LINKED ISSUE #321: `pydap` appears to be incompatible with `numpy>=2`
### Description

Importing `pydap.client` fails if `numpy>=2` is installed (noticed in `xarray`'s CI).

### Minimal example that reproduces the bug, or a new desired feature

With `numpy>=2` in the environment:

```pytb
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], line 1
----> 1 import pydap.client

File .../lib/python3.11/site-packages/pydap/client.py:53
     51 import pydap.lib
     52 import pydap.net
---> 53 import pydap.handlers.dap
     54 import pydap.parsers.dds
     55 import pydap.parsers.dmr

File .../lib/python3.11/site-packages/pydap/handlers/dap.py:16
     14 import re
     15 from itertools import chain
---> 16 from numpy.lib.arrayterator import Arrayterator
     18 # handlers should be set by the application
     19 # http://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library
     20 import logging

ModuleNotFoundError: No module named 'numpy.lib.arrayterator'
```

<details><summary>Environment</summary>

```
python : 3.11.9 (conda-forge)
numpy: 2.0.0
pydap: 3.4.0
```

</details>

Mikejmnez:
The following Pull Request:

- [x] Closes #321 
- [x] Tests are added to incorporate new behavior
- [x] Fixes docs and docstring  to display behavior with numpy 2 and Sequences.

### DIFF
--- a/.github/workflows/run_tests_macos.yml
+++ b/.github/workflows/run_tests_macos.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         pip3 install -U pip setuptools
         pip3 install -U "werkzeug<2.1"
-        pip3 install -U "numpy<2.0" webtest gsw coards
+        pip3 install -U numpy webtest gsw coards
         pip3 install -U netCDF4 pytest Webob Jinja2 docopt-ng beautifulsoup4 requests testresources
     - name: Run tests with pytest
       run: |
--- a/.github/workflows/run_tests_ubuntu.yml
+++ b/.github/workflows/run_tests_ubuntu.yml
@@ -25,7 +25,7 @@ jobs:
         python -m pip install --upgrade pip
         pip3 install -U pip setuptools
         pip3 install -U "werkzeug>=2.2.2" cryptography
-        pip3 install -U "numpy<2.0" webtest gsw coards
+        pip3 install -U numpy webtest gsw coards
         pip3 install -U netCDF4 pytest Webob Jinja2 docopt-ng beautifulsoup4 requests testresources
     - name: Run tests with pytest
       run: |
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -2,7 +2,7 @@ name: pydap_tests
 channels:
 - conda-forge
 dependencies:
-- numpy < 2.0
+- numpy
 - netCDF4
 - pytest
 - pytest-cov
--- a/docs/developer_data_model.rst
+++ b/docs/developer_data_model.rst
@@ -288,9 +288,9 @@ We can also iterate over the ``SequenceType``. In this case, it will return a se
 
     >>> for record in s.iterdata():
     ...     print(record)
-    (1, 10)
-    (2, 20)
-    (3, 30)
+    (np.int32(1), np.int16(10))
+    (np.int32(2), np.int16(20))
+    (np.int32(3), np.int16(30))
 
 Prior to pydap 3.2.2, this approach was not possible and one had to iterate directly over ``SequenceType``:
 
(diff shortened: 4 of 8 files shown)


Mikejmnez:
With Numpy = 2, a sequence now displays data slightly different. Consider the following example:
## Setup of example
```python
        >>> import numpy as np
        >>> data = np.array([
        ... (10, 15.2, 'Diamond_St'),
        ... (11, 13.1, 'Blacktail_Loop'),
        ... (12, 13.3, 'Platinum_St'),
        ... (13, 12.1, 'Kodiak_Trail')],
        ... dtype=np.dtype([
        ... ('index', np.int32), ('temperature', np.float32),
        ... ('site', np.dtype('|S14'))]))
        ...
        >>> seq = SequenceType('example')
        >>> seq['index'] = BaseType('index')
        >>> seq['temperature'] = BaseType('temperature')
        >>> seq['site'] = BaseType('site')
        >>> seq.data = data
```
###    Iteraring over the sequence to print data:

```python
        >>> for line in seq.iterdata():
        ...     print(line)
        (np.int32(10), np.float32(15.2), 'Diamond_St')
        (np.int32(11), np.float32(13.1), 'Blacktail_Loop')
        (np.int32(12), np.float32(13.3), 'Platinum_St')
        (np.int32(13), np.float32(12.1), 'Kodiak_Trail')

```
Note the `np.type` for each element of the array. 

### Before , the same code would print

```python
        (10, 15, 'Diamond_St')
        (11, 13.1, 'Blacktail_Loop')
        (12, 13.3, 'Platinum_St')
        (13, 12.1, 'Kodiak_Trail')
```
## Why?
With numpy 2 the following code throws an error

The change alone  introduced many errors. which required the change here
https://github.com/pydap/pydap/blob/e133372aea4f033b950c1137953dfd8b97c81b88/src/pydap/responses/dods.py#L45-L46

The error arises because newbytesorder is no longer a method of `x.astype` ...

Mikejmnez:
I found this simple way to decode the representation of numpy types into native Python types, with a slight modification to the following function. This would allow to revert all the changes to the documentation. That is, change the following function:
https://github.com/pydap/pydap/blob/e133372aea4f033b950c1137953dfd8b97c81b88/src/pydap/lib.py#L330-L335

To

```python
def decode_np_strings(numpy_var): 
    """Given a fixed-width numpy string, decode it to a unicode type""" 
    if isinstance(numpy_var, bytes) and hasattr(numpy_var, "tobytes"): 
        return numpy_var.tobytes().decode("utf-8") 
    else:
        if isinstance(numpy_var, np.generic):
            # starting with numpy 2.0 need to explicitly convert
            # numpy scalar to native Python type
            return numpy_var.item()
        else: 
            return numpy_var
```

However the new decode_np_strings yields now

```python
        >>> for line in seq.iterdata():
        ...     print(line)
    (10, 15.199999809265137, 'Diamond_St')
    (11, 13.100000381469727, 'Blacktail_Loop')
    (12, 13.300000190734863, 'Platinum_St')
    (13, 12.100000381469727, 'Kodiak_Trail')
```
instead of 
>
>        (10, 15, 'Diamond_St')
>        (11, 13.1, 'Blacktail_Loop')
>        (12, 13.3, 'Platinum_St')
>        (13, 12.1, 'Kodiak_Trail')
>
>

Nothing wrong - since this is the inherent precision of `np.float32`, but it may lead to the wrong interpretation that pydap is somehow changing the data.




### Thus
I decided NOT to change `np.decode_string` and instead change the documentation

Mikejmnez:
@jgallagher59701 ready for merger

jgallagher59701 (approved):
Ready to merge. Thanks!

Mikejmnez:
Awesome, thanks!

Issue #319, opened 2024-05-02
Resolve Deprecation Warnings 


Mikejmnez:
### Description
Running test with python 3.12, I noticed various `Deprecation Warnings`, one outside the scope of `pydap`, and the others within pydap. These are
### Within scope of pydap:
* **pkg_resources**. Largely resolved by #290, but care must be taken about the way namespaces are being discovered & imported with pydap (`pydap/handlers/lib.py`).
```python
src/pydap/handlers/lib.py:19
  ../pydap/src/pydap/handlers/lib.py:19: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources

../mambaforge/envs/pydap_tests/lib/python3.12/site-packages/pkg_resources/__init__.py:2832
  ../mambaforge/envs/pydap_tests/lib/python3.12/site-packages/pkg_resources/__init__.py:2832: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('paste')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)
```
* **for val in sequence**. This is internal within pydap >= 3.4. 
```python
  ../pydap/src/pydap/model.py:643: PendingDeprecationWarning: Starting with pydap 3.4 ``for val in sequence: ...`` will give children names. To iterate over data the construct ``for val in sequence.iterdata(): ...``is available now and will be supported in thefuture to iterate over data.
```

* **thread locks**. This is associated with fork vs spawn when using multiprocessing, as described in [pydap#292](https://github.com/pydap/pydap/issues/292). This is an important one to figure out.
```python
  ../mambaforge/envs/pydap_tests/lib/python3.12/multiprocessing/popen_fork.py:66: DeprecationWarning: This process (pid=30734) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

* **Conversion of an array with ndim > 0 to a scalar is deprecated** for numpy > 1.25
```python
  ../pydap/src/pydap/lib.py:161: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    return "%.6g" % obj
```

[...]

PR #325, opened 2024-07-15
Removes deprecation warnings

Mikejmnez:
The following Pull Request:

- [x] Partially closes an already existing opened issue #319. Many of these I have already resolved, but there were a couple of ones that were happening within pydap tests. 
- [x] Finishes and closes PR #289 started (thanks @facutuesca !): pyOpenSSL was deprecated but tests kept failing for macOS even though it wasn't explicitly being used. Turns out `pyOpenSSL` needed to be replaced with `cryptography` for tests to pass. See [pyca/pyopenss#1249](https://github.com/pyca/pyopenssl/issues/1249).  
on a local environment.

The reasoning behind muting Deprecation warnings is that these show up in many packages that have pydap as a dependency, for example [dfm_tools#803](https://github.com/Deltares/dfm_tools/issues/803), and that many of these Deprecation Warnings come from pydap internally (producing DeprecationWarnings, and failing to capture some other warnings).

Feel free to suggest a different approach to filtering these warnings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,7 +73,8 @@ tests = [
   'WebTest',
   'flake8',
   'werkzeug>= 2.2.2',
-  'gunicorn'
+  'gunicorn',
+  'cryptography'
 ]
 
 [tool.isort]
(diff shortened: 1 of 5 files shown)


Mikejmnez:
@jgallagher59701 ready for review...

jgallagher59701 (approved):
Ready to merge. Thanks.

Issue #326, opened 2024-07-17
remove GridType from DAP4 


Mikejmnez:
### Description

The DMR parser has this option:

https://github.com/pydap/pydap/blob/5e9bf297b0cf6a40bd2a4906a27600cda1cd2643/src/pydap/parsers/dmr.py#L200-L206

which creates a GridArray type. However, it is my understanding that GridArrays are no longer part of the DAP4 model. The way these GridArrays show up is when accessing data, say to plot or manipulate data.

### Minimal example

```python
from pydap.client import open_url
url = "dap4://oceandata.sci.gsfc.nasa.gov/opendap/PACE_OCI/L3SMI/2024/0301/PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc"
ds = open_url(url)
ds.tree()
```
returns
```python
.PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc
├──lat
├──lon
├──chlor_a
│  ├──chlor_a
│  ├──lat
│  └──lon
├──palette
├──rgb
└──eightbitcolor
```

where
```python
ds['chlor_a']
>>> <GridType with array 'chlor_a' and maps 'lat', 'lon'>
```
and so
```python
ds['chlor_a'][:]
```
downloads all `'chlor_a'`, `'lat'` and `'lon'`. 

### Proposed solution

To remove GridType arrays and thus allow for a much simpler dataset:
```python
ds.tree()
.PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc
├──lat
├──lon
├──chlor_a
├──palette
├──rgb
└──eightbitcolor
```

Mikejmnez:
closed by #327

PR #327, opened 2024-07-17
Simplify the Dataset model in DAP4

Mikejmnez:
The following Pull Request:

- [x] Closes an already existing opened issue #326  for removing GridType arrays from dmr parsers, thus there are no longer GridArrays in DAP4 datasets (there are simply arrays). This simplifies the Dataset model.
- [x] Tests are adjusted to fix the lack of GridType arrays.

In the example in #326, now:
```python
url = "dap4://oceandata.sci.gsfc.nasa.gov/opendap/PACE_OCI/L3SMI/2024/0301/PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc"
ds = open_url(url)
ds['chlor_a']
>>> <BaseType with data Dap4BaseProxy()...>
```
And ```ds.tree()``` yields:
```python
.PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc
├──lat
├──lon
├──chlor_a
├──palette
├──rgb
└──eightbitcolor
```

 A much simpler `pydap` dataset model in DAP4. 

Am I correctly interpreting one of many differences between DAP4 & DAP2 @jgallagher59701 ?

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -224,8 +224,6 @@ class DAPHandler(pydap.handlers.lib.BaseHandler):
                 session=self.session,
                 timeout=self.timeout,
             )
-        for var in walk(self.dataset, pydap.model.GridType):
-            var.set_output_grid(self.output_grid)
 
     def add_dap2_proxies(self):
         # now add data proxies
--- a/src/pydap/model.py
+++ b/src/pydap/model.py
@@ -260,7 +260,6 @@ class BaseType(DapType):
         # these are set when not data is present (eg, when parsing a DDS)
         self._dtype = None
         self._shape = ()
-        self._Maps = ()  # must be fully qualifying name
 
     def __repr__(self):
         return "<%s with data %s>" % (type(self).__name__, repr(self.data))
@@ -330,6 +329,7 @@ class BaseType(DapType):
         out.data = self._get_data_index(index)
         if type(self.data).__name__ == "BaseProxyDap4":
             out.attributes["checksum"] = self.data.checksum
+            out.attributes["Maps"] = self.Maps
         return out
 
     def __len__(self):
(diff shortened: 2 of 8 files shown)


jgallagher59701 (commented):
I'm not sure about this. Are you going to ignore the Map elements?

Mikejmnez:
> Are you going to ignore the Map elements?

So far, it is indeed being ignored. I guess the part that I am missing, is how are these `MAP` elements used in the DAP4 model? 

The way it was used it was to create `GridArrays` which are not part of the DAP4 model (if I understand correctly).

Mikejmnez:
Also - I noticed this got started some while ago:

https://github.com/pydap/pydap/blob/5e9bf297b0cf6a40bd2a4906a27600cda1cd2643/src/pydap/model.py#L893-L894

Don't know what the plan was with this.

Mikejmnez:
I think this is a better work around:

- [x] `GridType` is not used in DAP4. This is important because there is no copying of same data within the dataset. This means the following tree structure is preserved:

```python
.PACE_OCI.20240301_20240331.L3m.MO.CHL.V2_0.chlor_a.0p1deg.NRT.nc
├──lat
├──lon
├──chlor_a
├──palette
├──rgb
└──eightbitcolor
```

- [x] Maps is an attribute of a `BaseType` . In the example above the variable `chlor_a`contains two Maps: `lon` and `lat`. Before this PR `ds['chlor_a']` would be a GridType that would contain within itself copies of lon and lat (hence increasing the size of the dataset). Now:

```python
ds['chlor_a'].Maps
>>> ('/lat', '/lon')
``` 
which are fully qualifying names. And so this would work:
```python
import matplotlib.pyplot as plot
plt.contourf(ds[ds['chlor_a'].Maps[0]],ds[ds['chlor_a'].Maps[1]], ds['chlor_a'])
```

Mikejmnez:
Some last minute minor fixes

jgallagher59701 (approved):
Ready for merge. Thanks!

Issue #328, opened 2024-07-17
add numpy-like attributes to BaseType


Mikejmnez:
### Description
Numpy attributes such as [numpy.array.itemsize](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.itemsize.html) and [numpy.array.nbytes](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.nbytes.html), allow for lazy inspection of the memory that would be occupied by an array or entire dataset (uncompressed), if downloaded or fetched into a local interactive environment (or stored).

Currently, you need to retrieve data into memory as numpy arrays, and then you can estimate the memory usage (See example below). The attribute `nbytes` needs `shape` and `dtype`, which are already included in the DMR as metadata and thus already available information.

And so I think it would be very useful to have these numpy-like attributes in `BaseTypes`, to estimate array sizes when working with remote data that a user is considering working with in memory RAM (say on plot on a notebook), or downloading for local storage.

### Minimal example with present behavior

```python
import numpy as np
from pydap.client import open_url
url = 'dap4://test.opendap.org/opendap/data/nc/fnoc1.nc'
dataset=open_url(url)

dataset['u'].shape # already an attribute
>>> (16, 17, 21)
print('dtype: ', dataset['u'].dtype) # already an attribute
>>> dtype:  >i2
type(dataset['u'])
>>> pydap.model.BaseType

# Download data into numpy arrays
u = dataset['u'][:].data # this is now a numpy array
print('shape: ', u.shape)
>>> shape:  (16, 17, 21)
print('itemsize: ', u.itemsize)
>>> itemsize:  2
print('np.type: ', u.dtype)
>>> np.type:  int16
print('nbytes: ', u.nbytes)
>>> nbytes:  11424
```
### Minimal example with desired behavior

```python
dataset=open_url(url)

dataset['u'].itemsize # already an attribute
>>> 2
dataset['u'].nbytes
>>> nbytes:  11424

dataset.nbytes # compute `nbytes` for the entire dataset
>>> <int value here>
```

Mikejmnez:
closed by #329

PR #329, opened 2024-07-18
include numpy attributes to BaseType to compute arraysize in bytes (uncompressed)

Mikejmnez:
The following Pull Request:

- [x] Adds the feature described in #328
- [x] Tests are added to demonstrate the new behavior, and all tests pass
on a local environment.

With this PR one can do this:

```python
url = 'http://test.opendap.org:8080/opendap/podaac/LLC4320_pre-SWOT_Yongala_20121001.nc'

ds = open_url(url, protocol='dap4')
ds.nbytes
>>> 2448965660
```
This value is for uncompressed numpy array data. And so is much larger (almost 6x) than the value reported in the DAP Response Form: [445456145](http://test.opendap.org:8080/opendap/podaac/contents.html)

### DIFF
--- a/src/pydap/model.py
+++ b/src/pydap/model.py
@@ -175,7 +175,7 @@ from functools import reduce
 
 import numpy as np
 
-from .lib import _quote, decode_np_strings, tree
+from .lib import _quote, decode_np_strings, tree, walk
 
 __all__ = [
     "BaseType",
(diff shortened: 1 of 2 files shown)


jgallagher59701 (approved):
Ready to merge. Thanks.

PR #336, opened 2024-08-01
correctly parse projections (CEs) with Arrays in DAP4

### LINKED ISSUE #335: slices defined in a URL's Constraint expression are not being propagated correctly when retrieving remote data
### Description
Slices in the form of `[100:200]` and so on, defined in constraint expressions, affect arrays in both `DAP2` and `DAP4`. Broadly, internally when there is a slice, pydap always assumes it starts from the 0-index. This causes an error in `DAP2`, but there is no error in DAP4.

## Minimal example that reproduces the bug, or a new desired feature
```python
import matplotlib.pyplot as plt
import numpy as np

url = "http://test.opendap.org:8080/opendap/netcdf/examples/200803061600_HFRadar_USEGC_6km_rtv_SIO.nc"
```

### DAP4

```python
## get all longitude values
ds = open_url(url+"?dap4.ce=/lon")
lon = ds['lon'][:]
np.shape(lon)
>>> (701,)

## subset in DAP4 - get 100 points away from edge data values
ds_ce = open_url(url+'?dap4.ce=/lon[100:1:199]')
# URL of array is correct though!
ds_ce['lon'] 
>>> <BaseType with data Dap4BaseProxy('http://test.opendap.org:8080/opendap/netcdf/examples/200803061600_HFRadar_USEGC_6km_rtv_SIO.nc?dap4.ce=/lon[100:1:199]', 'lon', dtype('>f4'), (100,), (slice(None, None, None),))>
lon_ce1 = ds_ce['lon'][:].data
```
Even though the URL and the shape are correct,`slice(None, None, None)` contains no information about the slice in the CE.

The following comparisons FAILS but should be True:

```python
lon[100] == lon_ce1[0]
>>> np.False
assert (np.equal(lon[100:200], lon_ce1)).all()
>>> ---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Cell In[41], line 1
----> 1 assert (np.equal(lon[100:200], lon_ce1)).all()

AssertionError:
```

### DAP2
I get something slightly different, but the issue remains the same. Starting with the constraint expression:
```python
ds_ce = open_url(url+'?lon[100:1:199]', protocol='dap2')
# The following shows that there is a problem with the pydap-defined slice
ds_ce['lon']
>>><BaseType with data BaseProxy('http://test.opendap.org:8080/opendap/netcdf/examples/200803061600_HFRadar_USEGC_6km_rtv_SIO.nc', 'lon', dtype('>f4'), (100,), (slice(100, 100, 1),))>
```

The slice is defined as `slice(100, 100, 1)`. Even attempting to download data results an error:
```python
ds_ce['lon'][:].data
>>> HTTPError: 400 Bad Request
Error { 
    code = 400;
    message = "libdap error transmitting DataDDS: The start value of an array index is past the stop value.";
}
```

### proposed solution?

I think I know what is going on. THis is only an issue when the slicing defined in the Constraint Expression does NOT start from `0`. This is , somehow `pydap` fails to identify the start of the Slice. `pydap` correctly identifies the shape of the slice.

Also, repeating the exact steps described above, but with the following URL both DAP2 and DAP4 WORK!

```python
ds_ce = open_url(url+'?dap4.ce=/lon[0:1:99]')
```

Mikejmnez:
- [x] Closes #335 .
on a local environment.
- [x] Included within the testing, is the example described in #335 that was not working . 

Selections are not entirely being parse appropriately in DAP4. The work to enable that will be separate.

### DIFF
--- a/src/pydap/parsers/__init__.py
+++ b/src/pydap/parsers/__init__.py
@@ -13,7 +13,7 @@ from ..exceptions import ConstraintExpressionError
 from ..lib import get_var, unquote
 
 
-def parse_projection(input):
+def parse_projection(input, protocol="dap2"):
     """Split a projection into items.
 
     The function takes into account server-side functions, and parse slices
@@ -22,6 +22,10 @@ def parse_projection(input):
     Returns a list of names and slices.
 
     """
+    if protocol == "dap2":
+        key = ","
+    elif protocol == "dap4":
+        key = ";"
 
     def tokenize(input):
         start = pos = count = 0
@@ -30,7 +34,7 @@ def parse_projection(input):
                 count += 1
             elif char == ")":
                 count -= 1
-            elif char == "," and count == 0:
+            elif char == key and count == 0:
                 yield input[start:pos]
                 start = pos + 1
             pos += 1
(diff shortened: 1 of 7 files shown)


src/pydap/parsers/__init__.py
```diff
@@ -110,19 +114,20 @@ def parse_ce(query_string, protocol="dap2"):
...
-            projection = []
-            selection = tokens
-        else:
-            projection = parse_projection(tokens[0])
-            selection = tokens[1:]
-    else:
+    elif protocol == "dap4":
+        key = "|"
+        # remove `dap4.ce=` from query string
+        query_string = query_string[8:]
```
jgallagher59701, line 120: Does the code check that the 8 chars are actually _dap4.ce=_? It should, but if it's done elsewhere, that's good enough.

jgallagher59701 (approved):
Looks great but I have one question. If you test for that value, then merge, else add a test and if the code passes CI/CD, merge. No need for a second review.

Thanks.

Mikejmnez:
Added testing for catching scenarios with inconsistent specification of protocol, and syntax of CEs, things pass now. I will merge :)

PR #343, opened 2024-08-07
resolve Numpy>1.25 deprecation error

Mikejmnez:
The following Pull Request:

- [x] Solves one of two pending Deprecation warnings described on issue #319.
on a local environment.

### DIFF
--- a/src/pydap/lib.py
+++ b/src/pydap/lib.py
@@ -178,11 +178,10 @@ def encode(obj):
     if isinstance(obj, str) or isinstance(obj, np.ndarray) and obj.dtype.char in "SU":
         return '"{0}"'.format(obj)
 
-    # fix for DeprecationWarning: Conversion of an array with ndim > 0 to a 
+    # fix for DeprecationWarning: Conversion of an array with ndim > 0 to a
     # scalar is deprecated
     if isinstance(obj, np.ndarray) and np.ndim(obj) > 0:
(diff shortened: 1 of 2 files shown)


Mikejmnez:
This was a bit of a head scratcher... The test https://github.com/pydap/pydap/blob/463b73d5def243df1cea673c57273f52f27903f3/src/pydap/tests/test_handlers_lib.py#L411-L415

was resulting in a Deprecation warning related to encoding an array with `ndim>0`, as described in #319.


The function causing the deprecation warning was:

https://github.com/pydap/pydap/blob/463b73d5def243df1cea673c57273f52f27903f3/src/pydap/lib.py#L175-L184

 which can take arrays with elements that are strings, floats, etc. But when taking `IterData` in the testing, it was resulting a Deprecation warning. You can replicate the warning wiht the following snippet:

```python
print(["%.6g" % obj for obj in np.array([[2],[2]])])
>>> DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
```
You can fix the issue with:

```python
print(["%.6g" % obj[0] for obj in np.array([[2],[2]])])
````

But that approach breaks some other tests. I found a compromise with an extra conditional. Included the behavior with a tests.

Mikejmnez:
@jgallagher59701 ready for review :)

Mikejmnez:
"Approval not required" - will merge and continue working on one more deprecation error on separate PR

PR #346, opened 2024-08-09
Implicit discovery of entry points

### LINKED ISSUE #319: Resolve Deprecation Warnings

Mikejmnez:
The following Pull Request:

- Migrates away from the deprecated `pkg_resources` and adds `importlib-metadata` and `importlib-resources`, which remain compatible with Python 3.9. In the near future (when Python 3.9 is no longer actively maintained), we can drop these new dependencies. In theory we should be able to replace them with`importlib.metadata` and `importlib.resources` with very minimal syntax changes (`importlib is already included within `setuptools`, but changed behavior starting with Python 3.10).

Broadly,

- [x] Closes #319.
- [X] New Tests are not added but all tests pass on a local environment.

### DIFF
--- a/ci/environment.yml
+++ b/ci/environment.yml
@@ -9,6 +9,7 @@ dependencies:
 - pytest-env
 - codecov
 - scipy
+- pre-commit
 - pip
 - pip:
   - werkzeug >= 2.2.2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,8 @@ dependencies = [
   "docopt-ng",
   "beautifulsoup4",
   "lxml",
-  "requests"
+  "requests",
+  "importlib-metadata"
 ]
 description = "A pure python implementation of the Data Access Protocol."
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,8 @@ dependencies = [
   "beautifulsoup4",
   "lxml",
   "requests",
-  "importlib-metadata"
+  "importlib-metadata",
+  "importlib-resources"
 ]
 description = "A pure python implementation of the Data Access Protocol."
 dynamic = ["version"]
--- a/pytest.ini
+++ b/pytest.ini
@@ -8,6 +8,5 @@ markers =
     good_url
     auth
 filterwarnings =
-    ignore:.*:DeprecationWarning
     ignore:.*:UserWarning
     ignore:.*blockwise.*:UserWarning
--- a/pytest.ini
+++ b/pytest.ini
@@ -8,5 +8,6 @@ markers =
     good_url
     auth
 filterwarnings =
+    ignore:.*:DeprecationWarning
     ignore:.*:UserWarning
     ignore:.*blockwise.*:UserWarning
--- a/src/pydap/handlers/csv/__init__.py
+++ b/src/pydap/handlers/csv/__init__.py
@@ -9,8 +9,6 @@ import time
 from email.utils import formatdate
 from stat import ST_MTIME
 
-from pkg_resources import get_distribution
-
 from ...exceptions import OpenFileError
 from ...handlers.lib import BaseHandler, IterData
 from ...model import BaseType, DatasetType, SequenceType
@@ -106,7 +104,6 @@ class CSVHandler(BaseHandler):
         >>> temp_file.remove()
     """
 
-    __version__ = get_distribution("pydap").version
     extensions = re.compile(r"^.*\.csv$", re.IGNORECASE)
 
     def __init__(self, filepath):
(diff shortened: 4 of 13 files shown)


Mikejmnez:
A bit of context to the failed test on Ubuntu:

Test only fail for `Python 3.9`, and that is why these do not fail with MacOS. When I run these tests on OSX with Python < 3.10 tests also fail, but we already removed testing for OSX for Python<3.10 because of a changing behavior in GH workflows.

The reason test fail is because a change in behavior in the library `importlib.metadata` for Python >=3.10. 

One solution to this problem that I have seen around, is to try:

```python
try:
    from importlib.metadata import entry_points
except ImportError:  # Python < 3.10 (backport)
    from importlib_metadata import entry_points
```

However this does not work because `importlib.metadata` does not throw an `Exception` during `import`, and even if it did,`importlib_metadata` is not installed. `importlib_metadata` would need to be added either as a dependency for all python versions.

But I do not want to drop support/compat with Python 3.9, which is still one of the [stable versions](https://devguide.python.org/versions/) until 2025.


## Fix
For now, I will add `importlib_metadata` as a dependency to `pydap` until `Python 3.9` is no longer a stable version. This allows `pydap` to provide support for all stable Python versions until Python 3.13 comes along later in 2025.

jgallagher59701 (approved):
Ready to merge. Thanks!

Mikejmnez:
I just checked, and there are some (external) deprecation warnings for `Python >= 3.12` so I will suppress the `Deprecation Warnings` and merge :)

PR #348, opened 2024-08-14
correctly define `named dimensions` at `root` level

### LINKED ISSUE #347: AttributeError: '<class 'pydap.model.BaseType'>' object has no attribute '_is_string_dtype'
### Description

This error is confusing, because the object is indeed a `pydap.model.BaseType` and `_is_string_dtype` is a property of `BaseType`. This error only happens when trying to download data of a global dimension variable into memory, but the dataset creation goes smoothly.

### Minimal example
For this example, I am trying to access NASA data, and so it requires `token` authentication, read from a JSON file . The JSON file is created following the [GetStarted](https://github.com/OPENDAP/ESIP2024/blob/main/binder/GetStarted.ipynb) tutorial.

```python
import requests
import json
import requests

# load token json data
with open('token.json', 'r') as fp:
    Authorization = json.load(fp)

# pass Token Authorization to a new Session.
my_session = requests.Session()
my_session.headers = Authorization


# dataset parameters
filename = 'Daymet_Daily_V4R1.daymet_v4_daily_na_tmax_2010.nc'
hyrax_url = 'https://opendap.earthdata.nasa.gov/collections/C2532426483-ORNL_CLOUD/granules/'
url = hyrax_url + filename

## THIS DOES NOT WORK
ds = open_url(url, session=my_session, protocol='dap4')
ds.tree()
.daymet_v4_daily_na_tmax_2010.nc
├──y
├──lon
├──lat
├──time
├──x
├──tmax
├──lambert_conformal_conic
├──time_bnds
├──yearday
└──nv
```

The above works nicely. It is only when attempting to download data for the variable `nv` that I get the error:
```python
for child in ds.keys():
    print(child, ds[child].shape)
    if ds[child].ndim==1:
        ds[child][0]
    elif ds[child].ndim==2:
        ds[child][0,0]
    elif ds[child].ndim==3:
        ds[child][0,0,0]
```
yields:
```python
y (8075,)
lon (8075, 7814)
lat (8075, 7814)
time (365,)
x (7814,)
tmax (365, 8075, 7814)
lambert_conformal_conic ()
time_bnds (365, 2)
yearday (365,)
nv (2,)
AttributeError: '<class 'pydap.model.BaseType'>' object has no attribute '_is_string_dtype'
```

This means the variable `nv` is the one causing the trouble. For example:

```python
import pydap
print(isinstance(ds['nv'], pydap.model.BaseType)
>>> True
print(isinstance(ds['lon'], pydap.model.BaseType))
>>> True
print(ds['lon']._is_string_dtype)
>>> False
```
but
```python
try:
    ds['nv']._is_string_dtype
except:
    print('error')
>>> error
```

Interestingly enough, the variable `nv` is the only one that is listed as `Global Dimension`, but does not appear explicitly as a `Variable` in the Data Request Form (see [here](https://opendap.earthdata.nasa.gov/collections/C2532426483-ORNL_CLOUD/granules/Daymet_Daily_V4R1.daymet_v4_daily_na_tmax_2010.nc.dmr.html)). And so you cannot download even from the browser...

`nv` (and `niter`) is dimensions that are always `length=2` and can be defined as `np.array([0,1])`.

### solution

I think that since this variable cannot be downloaded, and it ALWAYS `ndim=1` and has `length=2`, it can be created locally. This would only be done when:
- it appears as a global dimension
- It does not appear as a Variable in the dataset
- ndim = 1
- len = 2

Mikejmnez:
The following Pull Request:

- [x] Closes #347 by setting named dimensions at root level.
- [x] Improves how dimensions are defined at root level in `Responses` (local creation of data --> DMR)  
- [x] Add tests. In particular that: a) A named dimension (e.g.`nv`) does not get assigned a variable/array, b) `Global dimensions` are appropriately parsed in the DMR and defined at root level, and c) simple exaamplt testing response (DMR generation from local dataset)

### DIFF
--- a/src/pydap/parsers/dmr.py
+++ b/src/pydap/parsers/dmr.py
@@ -216,8 +216,6 @@ def dmr_to_dataset(dmr):
         if "maps" in variable.keys():
             array.Maps = variable["maps"]
         var = array
-        if "skip_proxy" in variable.keys():
-            variable["attributes"]["skip"] = True
         var.attributes = variable["attributes"]
         if "parent" in variable.keys() and variable["parent"] in [
             "Sequence",
--- a/src/pydap/parsers/dmr.py
+++ b/src/pydap/parsers/dmr.py
@@ -162,6 +162,14 @@ def dmr_to_dataset(dmr):
     variables = get_variables(dom_et)
     named_dimensions = get_named_dimensions(dom_et)
 
+    # get Global dimensions at root level
+    global_dimensions = []
+    for name, size in named_dimensions.items():
+        if len(name.split("/")) == 1:
+            global_dimensions.append([name, size])
+
+    dataset.dimensions = tuple(tuple(item) for item in global_dimensions)
+
     # Add size entry for dimension variables
     for name, size in named_dimensions.items():
         if name in variables:
--- a/src/pydap/parsers/dmr.py
+++ b/src/pydap/parsers/dmr.py
@@ -154,9 +154,6 @@ def get_maps(element):
 def dmr_to_dataset(dmr):
     """Return a dataset object from a DMR representation."""
 
-    # this should simple be:
-    # return DMRParser(dmr).parse()
-
     # Parse the DMR. First dropping the namespace
     dom_et = DMRParser(dmr).node
     # emtpy dataset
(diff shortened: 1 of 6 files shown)


Mikejmnez:
@jgallagher59701 @ndp-opendap ready for review :)

PR #351, opened 2024-08-16
remove python 3.9

Mikejmnez:
This is a replica of another PR, which was idle-ing. 

The following Pull Request:

- [x] Removes Python 3.9
- [x] Tests pass locally


Mikejmnez:
Indeed the problem is when removing Python 3.9 (first commit) from github workflows python matrix,

Issue #349, opened 2024-08-15
Drop Support for Python 3.9


Mikejmnez:
### Description

As of April of this year, `Numpy` dropped support for `Python 3.9` (see [HERE](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule)). In our testing, we already dropped Python 3.9 for OSX, but have not for Ubuntu.

Mikejmnez:
This will wait until next version is released...

PR #356, opened 2024-08-22
Drop python 3.9

Mikejmnez:
The following Pull Request:

- [x] Closes issue #349. In doing so drops dependencies in favor of libraries within Python's standard library.
- [x] all tests pass locally

This is a 3rd try: See if with Admin access I can push this PR... (before I couldn't)


jgallagher59701:
So do wee need to look at the github workflows to address this?

Mikejmnez:
> So do we need to look at the github workflows to address this?

I am removing the python 3.9 from the `.github_workflows`. But somehow this PR is expecting some checking.. I read that closing and reopening the PR, and removing cache can help. But it didn't so far.

PR #369, opened 2024-09-11
drop docopt-ng, beautifulsoup4, lxml and others as required dependencies

### LINKED ISSUE #354: drop docopt-ng as required dependency
### Description
The original `docopt`, used by pydap, was long abandoned. `docopt-ng`, a forked of the original, is widely used and gets periodically maintained. Some months ago I replaced `docopt` for `docopt-ng` as a dependency, and it worked great as a replacement.

### issue
`docopt-ng` is only pip-installable and is NOT available on conda. This means that if we want to make a new release of pydap to be conda/mamba installable, we need to either 

a) convince `docopt-ng` maintainers to make it available on conda
b) drop `docopt-ng` as a dependency.

### solution
I ran
```python
grep -rnw src/pydap -e docopt
```
(`docopt` means `docopt-ng`) to inspect how/where docopt-ng is being used. Ther esults were:
```python
src/pydap/wsgi/app.py:26:from docopt import docopt
src/pydap/wsgi/app.py:283:    arguments = docopt(__doc__, version="pydap %s" % __version__)
```

This is easily replaceable. Since `docopt` is simply used by the app to interpret/parse arguments defined at the command line. We can replace it  with [argparse](https://docs.python.org/3/library/argparse.html), which already is part of the Python standard library.

Mikejmnez:
The following Pull Request:

- [x] Closes #354 and #349.
- [x] Tests are not added but pass on my local environment.

With regards to #349 -  I have disabled (temporarily) the workflow associated with running ubuntu tests. This is because I have attempted to drop python 3.9 many times and while all checks pass, github remains idle waiting for the ubuntu python 3.9 check even when the PR itself removes it. And so I am trying this new approach of disabling the entire ubuntu workflow to see if only running the macOS checks (plus pre-commit) allows for this PR to be merged onto main.

### DIFF
--- a/.github/workflows/run_tests_macos.yml
+++ b/.github/workflows/run_tests_macos.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install -U pip setuptools
-        pip install -e .[tests,netcdf,client]
+        pip install -e .[tests,netcdf]
 
     - name: Run tests with pytest
       run: |
--- a/.github/workflows/run_tests_ubuntu.yml
+++ b/.github/workflows/run_tests_ubuntu.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.x', '3.9', '3.10', '3.11']
+        python-version: ['3.x', '3.10', '3.11']
     name: ubuntu-latest Python ${{ matrix.python-version }}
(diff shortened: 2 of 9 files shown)


Mikejmnez:
@jgallagher59701 this PR is ready for review! Notice that I was able in this PR to finally remove Python 3.9 :)

jgallagher59701 (approved):
Ready to merge. Thanks.

PR #395, opened 2024-10-14
removes GridType from netcdf handler

### LINKED ISSUE #368: refactor to resolve hybrid DAP2/DAP4 behavior on netcdf handler when using as server
### Description
Pydap 's netcdf handler was updated to now serve nc4 data that contains groups (per this PR #367). But pydap's netcdf handler follows a hybrid DAP2/DAP4 model, to both allow data within Groups, while remain compatible with GridType. This results in

a) an inefficient code that, in order to generate dds, das and dmr responses, runs unnecessary extra parser code. That is, there is a parser for `dds`, another one for `das` and another one for `dmr` and these are largely independent from each other.

b) When serving nc4 dataset, the dmr can result in `attribute errors` when looking dimension variables that are defined at the `root` level (for example, when serving the [SimpleGroup](https://github.com/pydap/pydap/blob/8d36407c2b7b30cf29739faedefecbfa2f31762b/src/pydap/tests/datasets.py#L220-L257) dataset), resulting in incomplete DMR responses. This is because anything at the root level in the netcdf handler was left unmodified in PR #367 to remain compatible with the DAP2 data model.

## Solution
The big picture solution is that OPeNDAP wants to move towards the DAP4 as the default model implementation, and so the netcdf handler should be most consistent with DMRs. Moreover, to support DAP2 workflows, the `dds` and `dar` parsers should use the dmr parser internally to create the appropriate DAP metadata response. This implies a refactor of the `dds` and `das` parsers.

Mikejmnez:
The following Pull Request:

- [x] Closes #368, thus the internal pydap netcdf handler moves away from dap2 into pure dap4. The dmr response no longer needs a GridType to fully open a local nercd file. In addition, the handler got improved by using the local LazyVariable throughout. 
- [x] ~Tests are added to demonstrate the new or fixed behavior, and all tests pass
on a local environment.~ A test was removed, because despite passing a protocol='dap4' to the DapHandler, the `ServerSideFunction` did not recognize this and attempted to set up a dap2 contraint expression when subsetting the local file. This is a subject of a separate issue/PR (minimal priority).

### DIFF
--- a/src/pydap/handlers/netcdf/__init__.py
+++ b/src/pydap/handlers/netcdf/__init__.py
@@ -59,14 +59,17 @@ class NetCDFHandler(BaseHandler):
 
                 # Add dimensions when creating the DatasetType
                 Dims = OrderedDict()
+                fqn_dims = OrderedDict()  # keep track of fully qualifying names of dims
                 for dim in dims:
+                    fqn_dims.update({'/'+dim: dim})
                     if dims[dim] is None:
                         self.dataset.attributes["DODS_EXTRA"] = {
                             "Unlimited_Dimension": dim,
                         }
                     else:
-                        Dims.update({dims[dim].name: dims[dim].size})
+                        Dims.update({dims[dim].name: dims[dim].size}) 
                 # build dataset
+
                 name = os.path.split(filepath)[1]
                 self.dataset = DatasetType(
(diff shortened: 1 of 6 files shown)


jgallagher59701 (approved):
Ready to merge. Thanks.

Issue #401, opened 2024-10-22
Unable to download data from dap4 response generated by a TDS server


Mikejmnez:
### Description

Pydap can parse the metadata of remote files either DAP4 (`dmr`) or DAP2 (`dds`). However, I run into trouble when trying to download data on the fly when the data is hosted on a TDS server. "Downloading on the fly" really means: Passing a Contraint Expressions to the URL and request a dap response of only that subset. It is the server (TDS or Hyrax) that does all the work and pydap only communicates (I/O).
 
### My approach
I was curious so I downloaded the entire file (as netcdf) using the [Thredds HTTPService](https://thredds.ucar.edu/thredds/catalog/satellite/goes/east/products/SeaSurfaceTemperature/FullDisk/20241009/catalog.html?dataset=goes-east-products/SeaSurfaceTemperature/FullDisk/20241009/OR_ABI-L2-SSTF-M6_G16_s20242830000206_e20242830059514_c20242830104398.nc). Hyrax also has mature DAP4 implementation. 

With Hyrax there is no trouble downloading subsets of data on the fly. That last statement is not surprising because `pydap `has been actively developed in the past couple of years to parse dap4 responses from `test.opendap.org`, which is a hyrax server. 

### What does this mean?
There may be a small difference between the DAP4 response generated by a `TDS server`, and the DAP4 response generated by the `Hyrax server`. `Pydap` should be able to parse a DAP4 response regardless which server it is talking to.

### Minimal example

```python
url_tds= "https://thredds.ucar.edu/thredds/dap4/satellite/goes/east/products/SeaSurfaceTemperature/FullDisk/20241008/OR_ABI-L2-SSTF-M6_G16_s20242820000206_e20242820059514_c20242820104416.nc"
url_hyx = "http://test.opendap.org:8080/opendap/tds/OR_ABI-L2-SSTF-M6_G16_s20242820000206_e20242820059514_c20242820104416.nc"


dsT = open_url(url_tds, protocol='dap4')
dsH = open_url(url_hyx, protocol='dap4')

assert dsH.tree() == dsT.tree() # this is True
```
The above means metadata is the same (parsing dmr works as expected now). Now then downloading data on the fly:

### Hyrax
```python
dsH['y'][0:]
>>> <BaseType with data array([   0,    1,    2, ..., 5421, 5422, 5423], dtype=int16)>
```
### TDS
```python
dsT['y'][0:]
```
results in the full traceback:

```python
---------------------------------------------------------------------------
[...]

PR #415, opened 2024-11-18
Quick patchwork to parse TDS DAP4 responses

Mikejmnez:
This is a quick patchwork so that pydap CAN parse DAP4 responses from TDS.  See #401 

There are a couple of issues related to parsing DAP responses from a TDS. This PR solves:

a) Correctly split DMR from binary data in the response. There are very small differences in how the 2 servers split the dmr from the binary data:

### Hyrax
```
"... </Dataset>\n\r\n<binary data>"
```
### TDS
```
"... </Dataset>\r\n<binary data>"
```
DAP4 support within pydap has been written to accomodate Hyrax. The vary small difference between the two servers means that returning data from a TDS via DAP4 results in error.

With this PR, now there is no error but there is a "UserWarning" to inform users to use DAP2 protocol with TDS. The reason for that is this PR still does not fix another discrepancy between TDS and Hyrax servers. The discrepancy may be small, but relates to at least partially with how `endianness` is specified in the response.

For example, when reading from [this data_URL](http://test.opendap.org:8080/opendap/tds/OR_ABI-L2-SSTF-M6_G16_s20242820000206_e20242820059514_c20242820104416.nc.dmr.html) on Hyrax you CORRECTLY get:

```python
ds = open_url(data_url, protocol='dap4')
ds['y'][:]
>>> <BaseType with data array([   0,    1,    2, ..., 5421, 5422, 5423], dtype=int16)>
```
The same `data_url` running on TDS gets you:

```python
ds = open_url(data_url, protocol='dap4')
ds['y'][:]
UserWarning: Full DAP4 support for TDS data servers remains under development for pydap. We recommend to continue to use DAP2 protocol instead. See: https://github.com/pydap/pydap/issues/401
>>> <BaseType with data array([     0,    256,    512, ..., -26107, -25851, -25595], dtype='>i2')
```

The issue is NOT JUST endianness. And so for now passing a useful UserWarning is enough, while we spend some time working with the tds folks on resolving this issue.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -334,10 +334,10 @@ def safe_dmr_and_data(r, user_charset, url):
 
     dmr = dmr[4:] + b"</Dataset>"
     dmr = dmr.decode(get_charset(r, user_charset))
-    if 'thredds' in url.split('/') or 'dap4' in url.split('/'):
+    if "thredds" in url.split("/") or "dap4" in url.split("/"):
         i0 = 2
     else:
-        i0=3
+        i0 = 3
     data = data[i0:]  # for hyrax this is 3, for TDS this is 2 (not anymore?)
     return dmr, data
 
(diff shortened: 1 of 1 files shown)


Mikejmnez:
merging

PR #421, opened 2024-12-17
broaden dmr parser to allow multiple values on single attribute

### LINKED ISSUE #406: There are four ways to specify Attributes on DAP4 spec - Pydap should be able to parse all
### Description

The DAP4 specification allows for three ways to define an attribute (see [section 1.5.14](https://opendap.github.io/dap4-specification/DAP4.html#_attributes_and_arbitrary_xml) ):

1. Option 1 (tds does this)
```python
<Attribute name="name" type="{atomicTypeName|EnumType fqn}" value="value"/>
```

2. Option 2 (hyrax)
```python
<Attribute name="name" type="{atomicTypeName|EnumType fqn}"/>
  <Value>Value</Value>
</Attribute>
```

3. Option 3 (tds also does this)
```python
<Attribute name="name" type="{atomicTypeName|EnumType fqn}"/>
  <Value value="value"/>
</Attribute>
```

4. Option 4 
```python
<Attribute name="name" type="{atomicTypeName|EnumType fqn}">
  <Namespace href="http://netcdf.ucar.edu/cf"/> <!--optional-->
  <Value value="value"/>
  ...
  <Value value="value"/>
</Attribute>
```
Note that in this Option 4, multiple value are being defined for this single attribute, and that is valid scenario. What happens here then is that result should be a **vector of values** for that attributes.

Pydap can already, can parse attributes when there are defined following Options 1-3. But not Option 4

Mikejmnez:
~This is a draft~
The following Pull Request:

- [x] Closes  #406. Correctly parses an attribute with multiple values to generate a list of values.
on a local environment.

### example
This is included in the testing

```python
dmr = """
<Dataset name="foo">\n
<Int32 name="x">\n
    <Attribute name="attr" type="Float32">\n
        <Value></Value>\n
        <Value>2</Value>\n
        <Value value="1"/>\n</Attribute>\n
</Int32>\n
</Dataset>
"""
dom_et = DMRParser(dmr).node
dataset = DMRParser(dmr).init_dataset()
variables = get_variables(dom_et)
get_attributes(variables['x']['element'])['attr']
```
the last line returns the list for attribute `'attr'`. All values, excep None, are Float32, associated with variable `'x'` of type int32

```python
[None, 2.0, 1.0]
```

### DIFF
--- a/src/pydap/parsers/dmr.py
+++ b/src/pydap/parsers/dmr.py
@@ -85,22 +85,28 @@ def get_atomic_attr(element):
     Int_types = ["Int16", "Int32", "Int64", "int", "Int8"]
     uInt_types = ["uInt16", "uInt32", "uInt64", "uint", "uInt8", "Char"]
     name = element.get("name")
-    value = element.find("Value").text
-    if value is None:
+    value = [val.text for val in element.findall("Value")]
+    if value.count(None) > 0:
         # This could be because server is TDS.
         # If value is None still, then data is missing
-        value = element.find("Value").get("value")
+        indexes = value.index(None)
+        vals = [val for val in element.findall("Value")]
+        if isinstance(indexes, list):
+            for i in indexes:
+                value[i] = vals[i].get("value")
+        else:
+            value[indexes] = vals[indexes].get("value")
     _type = element.get("type")
-    if _type in dmr_atomic_types and value is not None:
+    if _type in dmr_atomic_types:
         if _type in Float_types:
             # keep float-type of value
-            value = float(value)
+            value = [float(val) for val in value if val is not None]
         elif _type in Int_types or _type in uInt_types:
             # keeps integer-type of value
-            value = int(value)
+            value = [int(val) for val in value if val is not None]
         else:
             try:
-                value = ast.literal_eval(value)
+                value = [ast.literal_eval(val) for val in value if val is not None]
             except ValueError:
                 # leaves value as string
                 raise Warning(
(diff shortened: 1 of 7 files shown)


Mikejmnez:
@jgallagher59701 ready for review

Mikejmnez:
actually - just realized that I am missing case 1 in #421 . Implementing this now.

Mikejmnez:
Done. Added plenty of testing and made sure all 4 cases (and combination of them) described in #406  are covered

src/pydap/parsers/dmr.py
```diff
@@ -89,22 +89,29 @@ def get_atomic_attr(element):
     Int_types = ["Int16", "Int32", "Int64", "int", "Int8"]
     uInt_types = ["uInt16", "uInt32", "uInt64", "uint", "uInt8", "Char"]
     name = element.get("name")
-    value = element.find("Value").text
-    if value is None:
-        # This could be because server is TDS.
-        # If value is None still, then data is missing
-        value = element.find("Value").get("value")
+    value = []
```
jgallagher59701, line 92: I think this code is hard enough to follow that you should add a comment or doc-string showing the different forms that attribute values can take in DAP4. For example, is the code on line 97 addressing a case like <Value>17 54 23</Value> which is an alternative to <Value>17</Value> <Value>54</Value> <Value>23</Value>?

Anyway, beyond some explanation, this looks great. It's a shame we wound up with such a disjoint specification, but that's what you get with 'design by committee.' ;-)

jgallagher59701 (approved):
Approved, but I think some more commentary will help us a year from now. ;-)

src/pydap/parsers/dmr.py
Mikejmnez: Added a docstring describing the 3 forms that attrs may be defined, and that these may be combined to defined a list of attribute values. I also refactored the code to extract values from 2 different forms at once.

Issue #432, opened 2025-02-03
Enable .netrc authentication


Mikejmnez:
One can define auth credentials persistently within a `.netrc` file, which http clients can parse.

The format for such netrc files is something like this:

```
machine example.com
login daniel
password qwerty
```

Pydap currently does not authenticate via `.netrc`, despite there being pythonic ways to do so. For example, consider the [netrc](https://docs.python.org/3/library/netrc.html) class. I think pydap should, so users do not have to define authentication parameters. In addiiton, tokens can be passed to a netrc file. This dual authentication can be of great use to simplify the code in pydap when authenticating, and make the documentation on authentication for general

Mikejmnez:
By default, the requests library reads the .netrc file and uses it for authentication. The PR #434 will make `pydap.net.GET` to use requests library, which will close this issue

PR #433, opened 2025-02-07
initial support for .netrc file for authentication

Mikejmnez:
**DRAFT**

Things to do before changing the status from draft.
- [ ] Enable authentication with tokens via the netrc file.
- [ ] Add tests for authentication with either hyrax / thredds / esfg. I checked, this tests are always skipped. In fact, the tests in tests/cas.urs usually fail. This is likely due to outdated workflow...

The following Pull Request:

- [ ] Closes an already existing opened issue #432  for fixing a bug, requesting
a new feature, etc.
- [ ] Tests are added to demonstrate the new or fixed behavior, and all tests pass
on a local environment.

### DIFF
--- a/src/pydap/cas/get_cookies.py
+++ b/src/pydap/cas/get_cookies.py
@@ -1,4 +1,5 @@
 import copy
+import netrc
 import warnings
 
 import requests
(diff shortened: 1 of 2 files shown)


Mikejmnez:
will close this, as no longer needed. With #434 , pydap.net.GET makes use of request library, which reads the .netrc by default.

PR #434, opened 2025-02-25
Drop Webob from client use of pydap

### LINKED ISSUE #429: modernize pydap's http client
### Description
pydap currently uses Webob for both the server and client. Webob is fine, but pydap could do better both on the server and client. The server and client aspect of pydap are largely decoupled (although we need to spend some time to make sure of this), and so we can make different choices for server and client. For the server the clear choice is FastAPI. However, there are a couple of choices for the client:

* Requests
* httpx
* aiohttp
* others?

We need to spend some time to consider the best alternative for pydap. In particular we should consider:

1. Async support
2. Speed
3. Stable dev
4. Ease of use
5.Authentication 

Other aspects to consider? Here are some basic resources
1. [Best Python HTTP Clients for Web Scraping in 2024](https://medium.com/@datajournal/best-python-http-clients-for-web-scraping-c09f7a356f29)
2. [Python Requests Alternatives](https://www.zenrows.com/alternative/python-requests)
3. [The best python http clients](https://www.scrapingbee.com/blog/best-python-http-clients/)

Mikejmnez:
The following Pull Request:
- [x] Replaces with use of Webob on pydap.net.GET in a general scenario where URL points to a remote (or local) dataset. Se below. Thus, this should close #429 and #432
- [x] Enable retries when requesting a GET via requests library on HTTP 500s. By default retries=5, but arguments can be passed directly to pydap.get_request()
- [x] Added more tests.
- [x] refactors `pydap.net` module. I remove several functions, in particular `pydap.net.raise_with_status` is no longer there. `pydap.net.follow_redirect` and `pydap.create_request` and `pydap.create_request_with_session` are combined together

In essence, this PR makes it so that 
- `pydap.net.GET` uses `requests.get(URL)` when the `application is None`, meaning that the URL points to a remote dataset. 
- `pydap.net` uses `Webob.Request.get_response` When application is not None, but intead is a wsgi applicatoin wrapping a pydap dataset. This case is mostly use for testing/development purposes.

### DIFF
--- a/src/pydap/cas/urs.py
+++ b/src/pydap/cas/urs.py
@@ -1,33 +1,19 @@
-from . import get_cookies
+from warnings import warn
 
 
 def setup_session(username, password, check_url=None, session=None, verify=True):
     """
-    A special call to get_cookies.setup_session that is tailored for
-    URS EARTHDATA at NASA credentials.
-
-    Parameters:
-    ----------
-    username: str
-    password: str
-    check_url: str, None (default)
-    session: `requests.session`, None (default)
-    verify: bool, True (default)
-
-    Example Data Access and Authentification in:
-    see `https://opendap.github.io/documentation/tutorials/`
-
+    This function is deprecated and will be removed in version 1.0.0.
+    Please use new_function instead.
     """
-
-    if session is not None:
-        # URS connections cannot be kept alive at the moment.
-        session.headers.update({"Connection": "close"})
-    session = get_cookies.setup_session(
-        "https://urs.earthdata.nasa.gov",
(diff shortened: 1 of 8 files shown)


jgallagher59701 (approved):
Lets merge this so you can move on. I'm not versed enough in this code to really review it, but I see you're using tests with mock objects. Do those tests simulate HTTP access?

Mikejmnez:
> Do those tests simulate HTTP access?

Yes indeed. You can create a redirect which is nice, and different types of errors. I used it to test that request was indeed following such redirect.

PR #438, opened 2025-03-05
improve requests.session workflow 

### LINKED ISSUE #437: Enable HTTP caching via requests library
By default, it looks like http caching is not a default behavior by the requests library. 

HTTP caching is important when accessing NASA data, since it can significantly reduce redirects to arrive to the final (signed) URL. As described by P. Quinn's comment:

> If you cache the redirect, then effectively you keep the final [CloudFront] URL, so if you’re peppering the file with requests, instead of hammering Lambda / S3 / CloudFront to get new presigned URLs then redirecting to those URLs, you just go straight to the original CloudFront URL until it expires (hours or days later)

To enable enable HTTP caching on the requests, you can use a [CacheControl](https://cachecontrol.readthedocs.io/en/latest/usage.html#adapter) via the [requests's transport adapter](https://docs.python-requests.org/en/latest/user/advanced/#transport-adapters). Pydap is [already using this Transport adapter to handle retries](https://github.com/pydap/pydap/blob/f9156d1f14beabeafc25d999660af306a9eba329/src/pydap/net.py#L142C1-L150C55), and so it would be easy to extend its use to handle caching for certain status codes.

Mikejmnez:
The following Pull Request:

- [x] adds [requests-cache](https://requests-cache.readthedocs.io/en/stable/index.html) as requirement to the Pyproject.toml, enabling the possibility of caching of opendap responses. 
- [x] when user requests a response, if the `session` has NOT been set by the user, pydap creates a session and that is persisted around the workflow of retrieving a dap response (see docstring of `pydap.net.create_session` for defaults when creating a session). Reusing this session is new behavior enabled by the PR. Before, the session was not passed and pydap's client had to continuously generate another session. Keep-alive is now the default of these session objects.
- [x] When user does not provide a session object, a new session object is created either using `requests.Session()` or `request_cache.CachedSession`. When one is used over the other depends on the user-defined parameter `use_cache`. `use_cache=False` by default, and that makes `requests.Session()` the default. 
- [x] user-defined `requests.get()`  keyword arguments are passed to the `daphandler`, allowing these to get reused when requesting dap responses (data). This gives additional control to the pydap user. 
- [x] When the client requests a dap response (binary data), `stream=True` is passed to  `request.get`. This is the recommender approach when requesting ["large" responses](https://requests.readthedocs.io/en/latest/user/advanced/#body-content-workflow).
- [x] When user passes a `token` for authentication, `pydap.get.create_session` uses it to update the session.headers, in according to the [documentation](https://pydap.github.io/pydap/notebooks/Authentication.html#token-approach).
- [x] `pydap.net.create_request` evaluates whether the session has a `Authorization` header, meaning a token, and configures the GET request to prevent recovery of username/password from a local `.netrc` file. 
- [x] Some tests are added to demonstrate correct arguments are being defined, when requesting a response from a server. More testing is neeed.

## What it needs to happen for this PR to be ready for review?

- [x] Decide how pydap uses requests-cache. ~So far, the default it set to False via either pydap.net.GET, or directly via get_responses. ~ I think this cache-behavior should NOT the default for now. 

### How to best use caching responses?

Whether pydap uses `CachedSession` for caching a response, or it does not cache a response, is set only once in the beginning of the pydap workflow. This is when creating a session object which will persist (keep-alive) the connection. In this phase only the dmr is requests to create the pydap object. The session is passed along, and eventually used to get dap responses (with constraint expressions).

An idea is to use CashedSession only when getting the dap response. However, in this scenario one would have to recreate a new Session, discarding the original session.


src/pydap/client.py
```diff
@@ -55,9 +55,11 @@
 import pydap.parsers.das
 import pydap.parsers.dds
 import pydap.parsers.dmr
-from pydap.handlers.dap import UNPACKDAP4DATA
+from pydap.handlers.dap import UNPACKDAP4DATA, DAPHandler
 from pydap.lib import DEFAULT_TIMEOUT as DEFAULT_TIMEOUT
 
+from .net import create_session
```
jgallagher59701, line 61: In this _the_ .net?

jgallagher59701 (approved):
Looks ready to go. I have one question.

src/pydap/client.py
Mikejmnez: this is `pydap.net`. In this scenario, it refers to the module net with `pydap` project. It is equivalent to

```python
from pydap.net import create_session
```

I am partial to `.net`, although pydap for the most part, uses the other alternative approach (you can see there on line 59 `from pydap.lib import ....` instead of `from .lib import ...`

jgallagher59701: OK, I get it.

Issue #442, opened 2025-03-08
400 errors attempting to access DAP4 data  from Thredds data with correct credentials


Mikejmnez:
The following test dataset https://thredds-test.unidata.ucar.edu/thredds/dap4/dev/d4icomp/restricted/SimpleGroup.nc4, which has a group, can be accessed and downloaded as long as the user is authenticated. However, even with the correct credentials (on a browser), accessing the `dmr.xml` (or attempting to download data) yields a 403 error. Refreshing the page allows one to proceed. This is a bug that is being corrected on the thredds side.

However, this retry-behavior can be successfully replicated with curl, but NOT with python's request library. Both libraries can handle redirects, but importantly requests library can also set retries. 

Figuring out how to access successfully dap4 data behind TDS with auth credentials is important, even after a fix from the TDS side, because not all servers get updated frequently.
 
### with curl
The following yields a status code 403, but stores the cookies on the ~/.cookies_file
```
curl -c ~/.cookies_file -b ~/.cookies_file -n -L "https://thredds-test.unidata.ucar.edu/thredds/dap4/dev/d4icomp/restricted/SimpleGroup.nc4.dmr"
```
Running the code above again works. It retrieves cookies data from the cookies file. 

### Issue with requests.session
Pydap uses a session object, which can discover credentials, update cookies, and handle redirects. So for example:

```python
import requests

session = requests.Session()
url = "https://thredds-test.unidata.ucar.edu/thredds/dap4/dev/d4icomp/restricted/SimpleGroup.nc4.dmr"
session.get(url, allow_redirects=True) 
>>> 400
```
Surprisingly, not only does it not work, but it yields a 400 error, and not a 403.

### closer look at curl
Running curl with `verbose`, I got:

```
> GET /thredds/restrictedAccess/opendapTesterRole HTTP/1.1
> Host: thredds-test.unidata.ucar.edu
> Authorization: Basic b3BlbmRhcFRlc3RVc2VyOm9wZW5kYXBUZXN0VXNlcg==
> User-Agent: curl/8.7.1
> Accept: */*
> Cookie: JSESSIONID=42B722B2981219B39E5AFF0C2D0EEEE0
> 
* Request completely sent off
< HTTP/1.1 403 403
HTTP/1.1 403 403
< Date: Fri, 07 Mar 2025 22:34:02 GMT
Date: Fri, 07 Mar 2025 22:34:02 GMT
< Server: Apache
Server: Apache
[...]

Issue #411, opened 2024-11-07
Can't get an authenticated connection to work to a THREDDS server


JimFluke:
I am trying to use authentication credentials to connect to our TDS. I have tried embedding the credentials into the url, but I get this error:
```
url: https://fluke:d1ef3ce7e7c41de74192a362524ad0a460692a222d9dd796ee383b56e446d749%241%24d03ce0f88475505a68bd0eb37fa570df8120e59ccf62a4f580a55ad612f695c0e385893fe7205f7c181b221ab49bc817d4a33a2b2bb727fdc0ee3420e7e5b99e@gcin01.cira.colostate.edu/thredds/dodsC/cloudsat-data/2B-GEOPROF.P1_R05/2008/366/2008366031107_14239_CS_2B-GEOPROF_GRANULE_P1_R05_E02_F00.hdf

Traceback (most recent call last):
  File "/app/opendap_pydap_two.py", line 64, in <module>
    dataset = open_url(url)
              ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pydap/client.py", line 68, in open_url
    handler = pydap.handlers.dap.DAPHandler(url, application, session, output_grid,
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pydap/handlers/dap.py", line 71, in __init__
    self.make_dataset()
  File "/usr/local/lib/python3.11/site-packages/pydap/handlers/dap.py", line 96, in make_dataset
    self.dataset_from_dap2()
  File "/usr/local/lib/python3.11/site-packages/pydap/handlers/dap.py", line 109, in dataset_from_dap2
    pydap.net.raise_for_status(r)
  File "/usr/local/lib/python3.11/site-packages/pydap/net.py", line 38, in raise_for_status
    raise HTTPError(
webob.exc.HTTPError: 401 Unauthorized
[...]

PR #443, opened 2025-03-08
patch to enable auth with tds + dap4 + authentication

Mikejmnez:
The following Pull Request:

- [x] A temporary, not very elegant patch to fix authentication issues with tds data server + DAP4 data. See #442 for context. Also #411  

### What is new with this PR

pydap can now access DAP4 data from a Thredds data server, when needing to authenticate. Authentication is handled by requests.Session object (via ~/.netrc), and pydap manually handles the cookies and redirect urls

The following now works, with authentication credentials on a local `~/.netrc` file.
```python
url = "https://thredds-test.unidata.ucar.edu/thredds/dap4/dev/d4icomp/restricted/SimpleGroup.nc4"
ds = open_url(url, protocol='dap4')
ds.tree()

.SimpleGroup.nc4
├──SimpleGroup
│  ├──Temperature
│  ├──Salinity
│  ├──Y
│  └──X
├──Pressure
├──time_bnds
├──time
└──Z


```

### NOTE
This is a temporary fix, while the bug is corrected on the server side.

### DIFF
--- a/src/pydap/net.py
+++ b/src/pydap/net.py
@@ -9,7 +9,7 @@ from requests_cache import CachedSession
 from urllib3 import Retry
 from webob.request import Request as webob_Request
 
-from .lib import DEFAULT_TIMEOUT, _quote
+from .lib import DEFAULT_TIMEOUT, _quote, __version__
 
 
 def GET(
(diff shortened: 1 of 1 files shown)


ndp-opendap:
From my perspective I think that this is really TDS dropping the ball. 
- Issuing/starting a new session after authentication is probably a bug. 
- Loosing the target URL so that the client can be correctly redirected to the requested resource is also a bug.,

Mikejmnez:
Closing this. No longer necessary as it seems this has been fixed on the tds end . See https://github.com/Unidata/tds/issues/575#issuecomment-3192766462

PR #477, opened 2025-04-11
Add Python3.13 metadata and fix spelling mistakes

Zeitsperre:
Hi there,

I use `pydap` on Anaconda, and I noticed that Python3.13 has recently received support here. I just wanted to be certain that the metadata reflects that change.

There also were a handful of small formatting, consistency, and spelling issues that I figured could be fixed at the same time.

Cheers,

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ dataset, and pydap will download the accessed data on-the-fly as needed. For exa
 ```python
     (1, 17999, 36000)
 ```
-> ![NOTE] 
+> ![NOTE]
 > In the example above, no data was downloaded, it was all lazily evaluated using OPeNDAP's DMR (DAP4) metadata representation. For more information, please check the documentation on [using pydap as a client](https://pydap.github.io/pydap/client.html).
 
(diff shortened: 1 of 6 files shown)


Mikejmnez:
Thanks @Zeitsperre! I did notice some of these spelling errors last week after I pushed (major) changes to the documentation. Truly appreciate the time you spent fixing these :) . 

BTW. Did you try building the documentation? Did the instructions make sense/ docs look fine? I changed things quite a bit, so any feedback on the matter would be great too!

FYI: I will add some more tutorials for best using pydap + xarray.open_mfdataset next week, so this is very timely, thanks!

Zeitsperre:
@Mikejmnez My pleasure! I hadn't built the documentation, but I just tried:

The console printout suggests a missing `references.bib` file and there are lots of `non-consecutive header level increase` and `document isn't included in any toctree` warnings but otherwise no significant errors! I can also see that both the English and Spanish directories have their own `environment.yml` files that mirror the one directly under `docs/` so it might be worthwhile to remove them and reference that one.

Addressing the warnings will probably take some effort, but it seems to working fine on my end!

Mikejmnez:
Merging, thanks again!

Issue #486, opened 2025-05-01
Improve safety practices on `consolidated_metadata`


Mikejmnez:
The `consolidate_metadata` function allows users to pre-download all dmrs associated with a datacube, and can assign a cache-key to a repeated dimensions when these dimensions appear across the dataset. The following are must to improve on the satefy of practices

- [x] Check that urls indeed form a datacube. This can be done by checking that N urls lead to N repeated variables and dimensions (and sizes?). 
- [x]  Allow users to specify a concat dimension. Most OPeNDAP datacubes have `time` as the concat dimension. So it is important not to assign the same cache key to all time dap responses. But lots of OPeNDAP urls do not have time as a dimension (it is part of their metadata). Hence this concat dimension must be optional

Mikejmnez:
closed by #488

PR #488, opened 2025-05-05
improvements to `consolidate_metadata`

Mikejmnez:
The following Pull Request:

- [x] Addresses #486 
- [x] Sets dims to return a list always #489 
on a local environment.

This PR will enable:

```python
consolidate_metadata(Temp_2017, cache_session, concat_dim='time', safe_mode=False, verbose=False)
```
where `Temp_2017` is a list with `N` URLs each pointing to a remote (cloud) dataset. `cache_session` is a requests-cache session object that can handle authentication.

The above line accomplishes:
 
* `safe_mode=False` (default) means ONLY a single DMR is downloaded. The rest of the DMRs (which may be 1000s) are assigned to a the same cache key as that of the one DMR.
* `concat_dim='time'` means that any dimension other than the `concat_dim` (`time`) is downloaded only once, with their dap responses of all other URLs are assigned to same cache key as the dap response of dimension on 1st url.
* For the `concat_dim='time'`, the dap response of this for each URL is downloaded and then cached.
* `verbose` enables to various printing as the cached keys are created for each dmr/dap response.

### Biggest Impact
Greatly reduces the Number of requests an OPeNDAP server receives, to generate an Xarray dataset. This is most significant for large N URLs, and when many dimensions are shared across the URL (datacubes - Level 3/4 data).

### Example usage

```python
from pydap.client import consolidated_metadata
from client.net import create_session

### if authentication is necessary, the snippet below assumes a .netrc file.
session=create_session(use_cache=True)
URLS = ["dap4://<rest_of_url1>", "dap4://<rest_of_url2>", ..., "dap4://<rest_of_urlN>"]
consolidated_metadata(URLS, session)

```


Mikejmnez:
# Benchmarks

### Dataset generation - No caching (i.e. old workflow)
Dataset generation refers to:
```python
xr.open_mfdatasetURLs[:N], engine='pydap', session=my_session, parallel=True, combine='nested', concat_dim='time', decode_times=True)
```
It creates a virtually aggregated view of the N-urls. cache_session is passed which contains token header for authenticatoin. There is a degree of parallelism involved in the background. But most of the time is spent downloading dap responses. 

Screenshot 2025-05-22 at 4 19 27 PM

Some points to take:
* It takes O(100)s of seconds to generate an xarray dataset consisting of N > 50 URLs.
* THis is a special dataset because there are 5 dimensions (all needed) within each urls (model output).
* Y-axes display same information: The time it took for Xarray+pydap to generate the dataset consisting of N opendap urls.
*  **Note the x-axes!!**: There is a huge difference between the N number of URLs the user defined, and the actual URLs that xarray+pydap downloaded to generate the Dataset!!!

Mikejmnez:
### Dataset Generation with caching (new enabled workflow)

now the code looks like:
```python
consolidate_metadata(Temp_urls[:N], cache_session, concat_dim='time')
xr.open_mfdataset(Temp_urls[:N], engine='pydap', session=cache_session,  parallel=True, combine='nested', concat_dim='time', decode_times=True)
```
The first step does the magic. The new info is the `concat_dim='time'` which before was not part of it.

Now there are two timings:
1. Consolidation of metadata. This is then cached and make available in a persitent way. Only needs to be executed once!
2. Xarray Dataset Generation. Because it reuses the cached-session. It is now I/O bound, no downloading of data takes place anymore to create the Xarray dataset.

Screenshot 2025-05-22 at 4 12 26 PM

Points to make:
* Black markers indicate timing to consolidated/cache all necessary metadata to create the Xarray Dataset. It is at least 10times faster. Only needs to be executed once.
*  Red markers. Time it takes to create an Xarray dataset. It takes a cached session (all metadata necessary is avaiable). It is 100x faster than creating the Xarray dataset without caching (right). 
* Creating the Xarray dataset without Caching. Same as before - for reference only.

docs/en/notebooks/ECCO.ipynb
```diff
@@ -1155,9 +1155,8 @@
    "id": "b67a4bbd-91d1-4aac-b23d-ea8a2d7cd9ae",
    "metadata": {},
    "source": [
-    "```{tip}\n",
-    "You can use the `consolidated_metadata method` to rapidly download all metadata associated with a datacube, and cache it.\n",
-    "This is a feature of pydap >= 3.5.5.\n",
+    "```{warning}\n",
+    "For `Pydap >= 3.5.5` the new method `consolidated_metadata` can be used to download and cache all dmrs in parallel. This can significantly speed up the (xarray) dataset exploration. THis method continues under active development and so its features will continue to evolve.\n",
```
jgallagher59701, line 1159: Minor typo: THis --> This. But maybe That last sentence could be: "However, this method remains under active..."

jgallagher59701 (approved):
This is really impressive! Please do post/share/shout about this!

The one caveat we have to make sure people understand is that the code uses the first DMR as a template and applies what it learns from that to all the remaining files/granules when it's run in the fastest mode (which is how most people are going to use it). This is the same design pattern that other have used, so it's a good choice, but  it can cause problems when people forget.

jgallagher59701:
I almost wish you could combine to the two plots above to show just how big the improvement is - the blue dots would run up high and the red dots much lower.

Mikejmnez:
> I almost wish you could combine to the two plots above to show just how big the improvement is - the blue dots would run up high and the red dots much lower.

hahaha yeah - I can combine them by adding multiple y-axes.

PR #493, opened 2025-05-19
set `https` as the default url scheme when dap2 or dap4 are used to set

Mikejmnez:
This PR makes the `https` the default url scheme to revert into, when a data user uses the URL scheme to define the DAP protocol.
### Step 1
User provides following url that specifies the DAP4 protocol:

```python
url = "dap4://<rest_of_data_url>" # pydap will download dmr
# or 
url = "dap2://<rest_of_data_url>" # pydap will download dds and das
```
### Step 2
Internally, `pydap` then sets `protocol="dap4"` (or `protocol='dap2'`) and will revertsthe scheme of the provided url to:

```python
url = "http://<rest_of_data_url>"
```
### Step 3:
pydap then uses the requests library to `GET` dmrs and dap responses. I have paid close attention to the redirects, and there is a 301 redirect (permanently moved) which then leads to the following url
```python
 "https://<rest_of_data_url>" # with relevant suffix appended here
```
This is usually the location of the data, behind an secure connection (`https`)

### New behavior:

This PR intends to avoid the 301 (permanently moved) url by setting the URL
```python
 `"https://<rest_of_data_url>"`
```
on Step 2.

### Compatibility with test.opendap.org

However, since the `test.opendap.org` data server only works with `http` and not `https` as of right now, this code enables to set `http` when the `base_url` is that of the test data server.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -121,7 +121,12 @@ class DAPHandler(BaseHandler):
         if self.scheme not in ["http", "https"]:
             if self.scheme in ["dap4", "dap2"]:
                 protocol = self.scheme
-                self.scheme = "http"  # revert to http
+                if self.netloc in ["test.opendap.org", "test.opendap.org:8080"]:
+                    # test.opendap.org is a special case
+                    # where https does not work
+                    self.scheme = "http"
+                else:
+                    self.scheme = "https"
                 return protocol
             else:
                 raise TypeError(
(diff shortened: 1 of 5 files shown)


jgallagher59701:
Can you have the code always try https first and then http if https fails?

Mikejmnez:
> Can you have the code always try https first and then http if https fails?

I can... I'll just have to look for a place for it with the precise exception ...

Mikejmnez:
@jgallagher59701  Done. All tests pass (except an outdated server test), and the performance appears unchanged  (it originally got slow due to automatic retries)

jgallagher59701 (approved):
Glad that worked! Looks ready to merge.

PR #509, opened 2025-06-13
Stream into temp file using a chunk manager when downloading dap responses

### LINKED ISSUE #508: improve memory management when downloading requests
When downloading requests via a session, the following is true:

```python
r = session.get(url, **extra_kwargs, stream=True)
```
However, when it gets to processing the dap response, the following remain from inherited code:

https://github.com/pydap/pydap/blob/0e182e66c6b59b01f1c158191d7a1e19537b294f/src/pydap/handlers/dap.py#L933

this loads the dap response into memory. Which is not ideal if the dap has O(Gbs) of data uncompressed. 

### Solution
Use a Chunk manager and write it (stream it) into a temporary file. python`s `tempfile` is a workable solution for this, and cross-platform. Essentially replace the line above with something that looks like:

```python
            with tempfile.TemporaryFile() as tmp:
                # write the response to a temporary file
                # so that we can read it in chunks
                for chunk in r.iter_content(chunk_size=CHUNK_SIZE):
                    if chunk:  # filter out keep-alive chunks
                        tmp.write(chunk)
                tmp.seek(0)
                self.raw = BytesReader(tmp)
```
where CHUNK_size is arbitrary, but since this will be used with dap responses, CHUNK_SIZE ~ (1) MB

Mikejmnez:
The following Pull Request:

- [x] Closes #508 . This would prevent downloading into memory dap responses.
- [ ] Tests are added to demonstrate the new or fixed behavior, and all tests pass
on a local environment.


Mikejmnez:
Failing tests also on main. See #510

Issue #510, opened 2025-06-13
Failing tests with string arrays


Mikejmnez:
this showed up on a PR, but it is also on main

```python
____________________________________________________________ TestArrayStringBaseType.test_getitem ____________________________________________________________

self = <pydap.tests.test_handlers_dap.TestArrayStringBaseType testMethod=test_getitem>

    def test_getitem(self):
        """Test the ``__getitem__`` method."""
>       np.testing.assert_array_equal(self.data[:], self.original_data)
E       AssertionError: 
E       Arrays are not equal
E       
E       Mismatched elements: 4 / 4 (100%)
E        ACTUAL: array([[b'This ', b'is '],
E              [b'a ', b'test']], dtype='|S5')
E        DESIRED: array([['This ', 'is '],
E              ['a ', 'test']], dtype='<U5')

src/pydap/tests/test_handlers_dap.py:646: AssertionError
================================================================== short test summary info ===================================================================
FAILED src/pydap/tests/test_handlers_dap.py::TestArrayStringBaseType::test_getitem - AssertionError: 
```

Mikejmnez:
partially addressed in #508 only runs that test if numpy < 2.3.0

Mikejmnez:
Fixed in #518

PR #517, opened 2025-06-30
String arrays

### LINKED ISSUE #516: add CRC32_checksum
Hyrax and Thredds include a flag in their dap responses (dap4) to compute checksums. These are added to the dap response as an element of the dmr portion, and the last 4 digits. pydap already checks for this element and parses it:

https://github.com/pydap/pydap/blob/98cdac8432358f4beda9e1daccfc8a08ae4e48d0/src/pydap/handlers/dap.py#L996-L1000

Both of the OPeNDAP servers implement CRC32_checksum. However, pydap does not compute and recheck that the value of the checksum matches that included in the dap response. 

Pydap should conpute this checksum and the user should be able to set this option (i.e. compute/evaluate checksums) an optional argument when using pydap (`False `by default)!

Mikejmnez:
The following Pull Request:

- [x]  Enables decoding dap responses of String arrays #518 
- [x] Improves handling of string arrays (improve compatibility with xarray). See issue #510  
on a local environment.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,5 @@
 [pytest]
-addopts = --doctest-modules --doctest-glob='*.rst'
+addopts = --doctest-modules
 doctest_optionflags= NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL ALLOW_UNICODE ALLOW_BYTES
 markers =
     server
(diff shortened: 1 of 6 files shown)


Mikejmnez:
@jgallagher59701 merging.

PR #519, opened 2025-07-01
Enable option to requests/skip checksums

### LINKED ISSUE #516: add CRC32_checksum

Mikejmnez:
The following Pull Request:

- [x] Closes #516 
on a local environment.


Mikejmnez:
@ndp-opendap @jgallagher59701 
One of the things I am struggling with is that I am seeing no difference whether I append `dap4.checksums=true|false` to the dap url. It seems I get the same response.

### Example

```python
import requests
session = requests.Session()

r_true = session.get("http://test.opendap.org/opendap/dap4/SimpleGroup.nc4.h5.dap?dap4.ce=/SimpleGroup/Y%5B0:1:39%5D&dap4.checksum=true")

r_false = session.get("http://test.opendap.org/opendap/dap4/SimpleGroup.nc4.h5.dap?dap4.ce=/SimpleGroup/Y%5B0:1:39%5D&dap4.checksum=false")

# test for differences between the two:
assert t_true == t_false
>>> True
```
So no difference in the two responses...

ndp-opendap:
There are no differences because test.opendap.org is ignoring the `dap4.checksum` query string parameter and always returning checksums :)

Mikejmnez:
> There are no differences because test.opendap.org is ignoring the `dap4.checksum` query string parameter and always returning checksums :)

OK! I missed the part about ignoring the checksum query string parameter...

PR #525, opened 2025-07-10
Enables batching multiple variable requests into a single dap url

Mikejmnez:
The following Pull Request:

- [x] Closes #521 .
- [ ] Tests are added to demonstrate the new or fixed behavior, and all tests pass
on a local environment.

### Minimal Example
```python
from pydap.client import open_url
from pydap.net import create_session

url = "http://test.opendap.org/opendap/dap4/SimpleGroup.nc4.h5"
session = create_session(use_cache=True)
session.cache.clear()

pyds = open_url(url, protocol='dap4', session=session)
pyds.tree()
.SimpleGroup.nc4.h5
├──SimpleGroup
│  ├──Y
│  ├──X
│  ├──Temperature
│  └──Salinity
├──time
├──Z
├──Pressure
└──time_bnds


```

Now the new stuff
```python
# Enable batch mode. Set the timeout for 1 second. This is any data request that together take place within 1 second will
# be processed together within a single dap url

pyds.enable_batch_mode(timeout=1)


# trigger download subset of 2 variables at once

temp = pyds['SimpleGroup/Temperature'][0,10:20,10:20].data;
salt = pyds['SimpleGroup/Salinity'][0,10:20,10:20].data;

```

check that there is a single dap url by looking at the cached url

```python
session.cache.urls()
>>> ['http://test.opendap.org/opendap/dap4/SimpleGroup.nc4.h5.dap?dap4.ce=%2FSimpleGroup%2FSalinity%5B0%3A1%3A0%5D%5B10%3A1%3A19%5D%5B10%3A1%3A19%5D%3B%2FSimpleGroup%2FTemperature%5B0%3A1%3A0%5D%5B10%3A1%3A19%5D%5B10%3A1%3A19%5D',
 'http://test.opendap.org/opendap/dap4/SimpleGroup.nc4.h5.dmr']
```

Lastly, check that the data is processed correctly:

```python
temp[:]
>>> array([[[410., 411., 412., 413., 414., 415., 416., 417., 418., 419.],
        [450., 451., 452., 453., 454., 455., 456., 457., 458., 459.],
        [490., 491., 492., 493., 494., 495., 496., 497., 498., 499.],
        [530., 531., 532., 533., 534., 535., 536., 537., 538., 539.],
        [570., 571., 572., 573., 574., 575., 576., 577., 578., 579.],
        [610., 611., 612., 613., 614., 615., 616., 617., 618., 619.],
        [650., 651., 652., 653., 654., 655., 656., 657., 658., 659.],
[...]

PR #527, opened 2025-07-17
Enable unpacking responses from httpx

Mikejmnez:
The following Pull Request:

- Enables, if installed, the use of `httpx` to download dap responses. 
 

### Why?

`httpx` is:
- Modern python http client (more than requests)
- Async / http/2 support
- Almost identical syntax to Requests

`httpx` will NOT be listed as a dependency (not even optional) but if installed and if the response is from a `httpx` client, the class `UNPACKDAP4DATA` can decode the `dap response` and return a correct `BaseType` data array

With some few download testing I have done, the first download using the `httpx` client is 1-3x faster than using `requests`. At worst, it the the same speed, but can be significantly faster.

This small change in the code will enable further benchmarking download timings on PR #525

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -49,6 +49,11 @@ from pydap.parsers.dds import dds_to_dataset
 from pydap.parsers.dmr import dmr_to_dataset
 from pydap.responses.dods import DAP2_response_dtypemap
 
+try:
+    import httpx
+except ImportError:
+    httpx = None
+
 logger = logging.getLogger(__name__)
 logger.addHandler(logging.NullHandler())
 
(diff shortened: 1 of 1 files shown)


Mikejmnez:
# Example
The following works now

```python
import httpx
from pydap.handlers.dap import UNPACKDAP4DATA

client = httpx.Client(follow_redirects=True, timeout=20, headers={"Authorization": "Bearer <TokenHere>"})
r = client.get(single_dap_url)
dataset = UNPACKDAP4DATA(r).dataset

dataset['time'][:].data
>>> array([   0,   60,  120,  180,  240,  300,  360,  420,  480,  540,  600,
        660,  720,  780,  840,  900,  960, 1020, 1080, 1140, 1200, 1260,
       1320, 1380], dtype=int32)
```

PR #530, opened 2025-07-18
Ensure `consolidate_metadata` and `batch_mode` are well integrated

### LINKED ISSUE #529: Make sure Consolidate Metadata integrates well with batching mode
`pydap` `consolidate metadata` enables a fast aggregation of multiple pydap datasets that share all but one dimension, the later being a concat dimension. Internally, it works as follows:

### how does consolidate_metadata works?

1. User specifies a list of urls (must share identical base_url), and specifies the dimension along which the datasets concatenate (e.g. `concat_dim='time'`). 
2. Downloads and caches the metadata for all urls.
3. Downloads all dimensions except the `concat_dim` only once. NOTE: each dimension is downloaded separately on its own dap url
4. Downloads N-dap responses of the `concat_dim`
5. All downloads are cached within the same session, resulting in a static metadata object that enables dataset aggregation of N datasets in O(1) second.

### What's new with `enable_batch_mode`?

With the merger of #525, multiple dimensions can be downloaded with a single dap url (as they should be), all streamed together. This can speeds things up to `10x` the initial dataset generation. However, this new enhancement feature needs to integrate better with consolidate metadata.

### Potential issues

Consolidate metadata cached urls of dimensions in the followng approach:

1. Non concat_dim urls:
```python
https:/<base_url>.dap?dap4.ce=/Var1[slices()];/Var2[slices()];/Var3[slices()];...;VarN[slices()]
```
Where the ordering of `{Var1, Var2, ..., VarN}` matters. If there is a discrepancy, the session object will then download the entire thing again and defeats the purpose of this approach.
2. Creates a separate download for the `concat_dim`.
```python
https:/<base_url>1.dap?dap4.ce=/concat_dim[slice]
https:/<base_url>2.dap?dap4.ce=/concat_dim[slice]
...
https:/<base_url>M.dap?dap4.ce=/concat_dim[slice]
```

This bring the potential issues:

a) When downloading multiple variables at once, the ordering of how these variables appear is not fixed and therefore there is no guarantee that it will match the cached urls.

b) Must filter `concat_dim` out of the batching, because this dimension (usually `time`) is downloaded separately. And so `Consolidate_Metadata` must provide a keyword or something to pass along, and that can be used to filter this dimension array.

Mikejmnez:
…r discovery

The following Pull Request:

- [x] Closes #529 , by passing along a `concat_dim` parameter to the session header, and ensure correct ordering of variables that appears in the dap url.
on a local environment.

# Example downloading MERRA-2 data
* Uses the CMR to only access the granules from Jan/1/2023 --- Jan/31/2023 
(1 month, 31 granules, each granule has hourly data)

* Using Hyrax to subset before download: all lons,`-80 < lats < -45`
```python


# code not shown queries the CMR for urls, and use constraint expressions to only keep `["/T2M", "/U2M", "/V2M", "/SLP"] ` and their dimensions arrays

len(opendap_urls)
>>> 31

%%time
consolidate_metadata(opendap_urls, concat_dim='time', session=my_session)
>>> datacube has dimensions {'lon[0:1:575]', 'lat[0:1:360]'} , and concat dim: `time`
CPU times: user 530 ms, sys: 163 ms, total: 694 ms
Wall time: 16.5 s

%%time
ds = xr.open_mfdataset(opendap_urls, engine='pydap', session=my_session, combine='nested', concat_dim="time", chunks={'lat':90}, batch=True)
>>> CPU times: user 557 ms, sys: 75 ms, total: 632 ms
Wall time: 704 ms


# lazily compute the daily mean - does not trigger computation 
%%time
nds = ds.isel(lat=slice(20, 90))
nds = nds.resample(time="1D").mean()
>>> CPU times: user 158 ms, sys: 2.01 ms, total: 160 ms
Wall time: 161 ms

######### Store subset data in a local `Test.nc4` file. 
# Before storing, the code below triggers all computations that were "lazy". This is, downloads dap responses of all `T2M`, `U2M`, `V2M` and `SLP` (31 granules, 1 dap url downloads all data for that granule). And before writting into diwnk, xarray computes the time average (daily) per variable .

%%time
nds.to_netcdf("Test.nc4", mode='w')
>>> CPU times: user 3.03 s, sys: 2.26 s, total: 5.29 s
Wall time: 35.3 s


# check that the dap response of a granule is consistent:
my_session.cache.urls()[5].replace("%5B", "[").replace("%5D","]").replace("%3A",":").replace("%3B",";")
>>> 'https://opendap.earthdata.nasa.gov/collections/C1276812863-GES_DISC/granules/M2T1NXSLV.5.12.4:MERRA2_400.tavg1_2d_slv_Nx.20230103.nc4.dap?dap4.ce=SLP[0:1:23][0:1:89][0:1:575];T2M[0:1:23][0:1:89][0:1:575];U2M[0:1:23][0:1:89][0:1:575];V2M[0:1:23][0:1:89][0:1:575]&dap4.checksum=true'

```

The CE: `dap4.ce=SLP[0:1:23][0:1:89][0:1:575];T2M[0:1:23][0:1:89][0:1:575];U2M[0:1:23][0:1:89][0:1:575];V2M[0:1:23][0:1:89][0:1:575]&dap4.checksum=true`

shows that indeed Hyrax got the request to slice all arrays in the latitude dimension (`[0:1:89]`), and that the cached URL contains all variables in it (as opposed to, before, a single dap url per variable)

### without batch=True

The last step takes 1min 30 second

### DIFF
--- a/src/pydap/client.py
+++ b/src/pydap/client.py
@@ -121,9 +121,9 @@ def open_url(
         is 'dap4'. If the URL ends with '.dods', the protocol is 'dap2'. Another
         option to specify the protocol is to use replace the url scheme (http, https)
         with 'dap2' or 'dap4'.
-    batch: bool (Default: True)
-        Flag that indicates download multiple arrays with single dap response. Not true
-        before.
+    batch: bool (Default: False)
+        Flag that indicates download multiple arrays with single dap response. Only
+        compatible with DAP4 protocol.
     use_cache : bool (Default: False)
         Whether to use the cache or not in the requests.
     session_kwargs: dict | None
--- a/src/pydap/lib.py
+++ b/src/pydap/lib.py
@@ -369,6 +369,26 @@ def decode_np_strings(numpy_var):
 
 
 def register_all_for_batch(ds, dims, concat_dim=None):
+    """
+    Used to register all dimension array when pydap
+    dataset has been initialized with batch=True.
+
+    Parameters:
+    ----------
+        ds: pydap dataset
+            Must be DAP4, and remote data.
+        dims: list
+            List of dimensions in `ds` that will be processed
+        concat_dim: str | None (default)
+            name of concatenating dimension (e.g. `time`). If not `None`
+            this dimension will NOT be processed together with the other
+            dimensions
+
+    Returns:
+        BatchPromise object
+            all arrays get `registered` to be downloaded together within
+            single dap response
+    """
     promise = ds._current_batch_promise = BatchPromise()
     for name in dims:
         if name in ds.keys():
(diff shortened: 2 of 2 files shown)


jgallagher59701 (approved):
This looks good, but  wonder, should batch mode be false by default? It looks like it is set to False on line 89 of client.py. But, then on 124 it looks like it's True by default.

It might just be the diffs are hard to read, but I thought I'd mention it.

PR #574, opened 2025-10-01
Enable batch=False|True when consolidating metadata

### LINKED ISSUE #573: enable consolidating_metadata with batch=True|False
Right now, `consolidate_metadata` downloads all dimensions within single dap url, same for the Maps. This is only helpful IFF xarray.open_dataset supports batch=True. There is a PR for that, but it may take a little bit of time to get that merged. And even when that happends, that will likely not feature until the next release.

In the meantime, we can allows batch as an optional parameter too. If False (the default) is should download each dimensions with its own dap url. If True, do it the way it does not. That will support current versions of xarray, and hopefully eventually we will be able to make batch=True as default and deprecate this argument.

### LINKED ISSUE #575: When consolidating metadata, make stronger check on dimensions
the method `pydap.client.consolidate_metadata` makes a safe check that dimensions match by name across all dmrs. THe snippet is here:

https://github.com/pydap/pydap/blob/1871b9f3125058cc04cef8c42f5283950bf68b61/src/pydap/client.py#L265-L275

`open_dmr(url)` simply creates an empty dataset that only contains the dimensions. For example, 

```python
{'tile': 13,
 'j_g': 90,
 'i_g': 90,
 'k_p1': 51,
 'k_l': 50,
 'j': 90,
 'i': 90,
 'time': 1,
 'k': 50,
 'nb': 4,
 'k_u': 50,
 'nv': 2}
```

The check in the snippet above, only compares the dimension names. To make this a stronger check, it should compare the dimension sizes too.

Mikejmnez:
The following Pull Request:

- [x] Closes #573 , setting batch=False as default (until the PR on xarray is merged, this is the alternative).
- [x] Closes #575  
on a local environment.

@jgallagher59701  Ready for review. The following example works well:

### example

consider the following list of dap4 urls of the form
```python
dap4://opendap.earthdata.nasa.gov/providers/POCLOUD/collections/ECCO%20Ocean%20Temperature%20and%20Salinity%20-%20Monthly%20Mean%20llc90%20Grid%20(Version%204%20Release%204)/granules/OCEAN_TEMPERATURE_SALINITY_mon_mean_1999-12_ECCO_V4r4_native_llc0090?dap4.ce=/THETA;/SALT;/k;/tile;/j;/i;/time
```
This granule has 5 dimension, and consider the case where I want to aggregate a list of 217 such granules along the time dimension. So URLS is a list of 217 elements

```python
%%time
consolidate_metadata(URLs, concat_dim='time', session=session) # batch=False by default
>>> CPU times: user 2.66 s, sys: 955 ms, total: 3.62 s
Wall time: 34.1 s
```
It downloads all dms, the time dimension for each granule, and the rest of the dimensions but for only a single granule . These are:
```
non_concat_dims = ['tile', 'k', 'i', 'j']
```

NOTE that `batch=False` implies that each dimension gets its own dap request.

Now, I use xarray to aggregated these in parallel:
```python
%%time
ds = xr.open_mfdataset(
    Temp_2017,
    engine='pydap',
    session=session,
    parallel=True,
    combine='nested',
    concat_dim='time',
)
>>> CPU times: user 7.58 s, sys: 1.25 s, total: 8.83 s
Wall time: 32.6 s
```

Screenshot 2025-10-01 at 11 10 21 PM

This means that `session`, which is a CachedSession object correctly uses the cache_key generated in `consolidate_metadata`, to avoid repeated data N times.

### DIFF
--- a/src/pydap/tests/test_client.py
+++ b/src/pydap/tests/test_client.py
@@ -1323,3 +1323,26 @@ def test_data_check(var_batch, key_batch, var_name, var_key, expected_shape):
     data = data_check(var, var_key)
 
     assert data.shape == expected_shape
+
+
+def test_consolidate_metadata_non_batch(cache_tmp_dir):
+    """Test that consolidate_metadata raises an error when batch=False"""
+    urls = [
+        "dap4://test.opendap.org/opendap/hyrax/data/nc/coads_climatology.nc",
+        "dap4://test.opendap.org/opendap/hyrax/data/nc/coads_climatology2.nc",
+    ]
+    cache_name = cache_tmp_dir / "test_consolidate_metadata_non_batch"
+    cached_session = create_session(
+        use_cache=True,
+        cache_kwargs={"cache_name": cache_name},
+    )
+    cached_session.cache.clear()
+    consolidate_metadata(
+        urls,
+        session=cached_session,
+        concat_dim="TIME",
+        batch=False,
+    )
+
+    assert cached_session.settings.key_fn._concat_dim == ["TIME"]
+    assert cached_session.settings.key_fn._collapse_vars == {"COADSX", "COADSY"}
--- a/src/pydap/tests/test_client.py
+++ b/src/pydap/tests/test_client.py
@@ -1348,6 +1348,26 @@ def test_consolidate_metadata_non_batch(cache_tmp_dir):
     assert cached_session.settings.key_fn._collapse_vars == {"COADSX", "COADSY"}
 
     N = len(urls)  # N of DMRS
-    N_repeated_dims = 2  # COADSX, COADSY
+    N_non_concat_dims = 2  # COADSX, COADSY
     N_concat_dims = len(urls)  # TIME
-    assert len(cached_session.cache.urls()) == N + N_repeated_dims + N_concat_dims
+    assert len(cached_session.cache.urls()) == N + N_non_concat_dims + N_concat_dims
+
+    # ============ ONLY WORKS FOR CLOUD OPENDAP URLS - SKIPPED FOR NOW
+    # # check that all URLS from COADSX and COADSY are cached, even when only 1 of each
+    # # was downloaded
+    # map1 = _quote("/COADSX[0:1:179]").replace("/", "%2F")
+    # map2 = _quote("/COADSY[0:1:89]").replace("/", "%2F")
+
+    # dap_urls = [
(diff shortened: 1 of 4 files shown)


src/pydap/tests/test_client.py
```diff
@@ -1308,3 +1323,50 @@ def test_data_check(var_batch, key_batch, var_name, var_key, expected_shape):
     data = data_check(var, var_key)
 
     assert data.shape == expected_shape
+
+
+def test_consolidate_metadata_non_batch(cache_tmp_dir):
```
Mikejmnez, line 1328: @jgallagher59701 This is the test that matters.

Both files have the same identical dimensions COADSX and COADSY. The method `consolidate_metadata` creates a special key so that these are downloaded from only 1 granule (separte dap url). And for any additional URL that has the same COADSX dimension and has a nearly identical url, it reuses the cached response for COADSX .


The last assertion proofs that. Even then the dap url is not on the cache, it reads data from the cache.

jgallagher59701: This is a tricky optimization - because it could be 'fooled' by badly  formed data. But it fits with some much of the multi-file data we see that it makes sense. This was used in NcML and the TDS when it did/does aggregations and also our NcML handler in Hyrax, although in those cases the data provider set that up through configuration.

Is this a behavior that can be turned on/off by the user/client?

Mikejmnez: Yes - if a user skips the `consolidate_metadata` step, and simply do:

```python
ds = xr.open_dataset(...)
```
then none of the "optimization" takes place. 

And restarting a session (for example, when restarting a jupyter kernel) also removes all of these optimization

```python
session = CashedSession(cache_name='name_to_db')
```
that will restart the session, pointing to the databaset, but without any of the cache_key optimization. It will be significantly slow still, since xarray will download all the dimensions for each granule (which are not stored originally in the database)

Mikejmnez:
Added a strong assertion check that dimensions and their sizes must match across all DMRs. Otherwise consolidate_metadata exists with a warning.

src/pydap/tests/test_client.py
jgallagher59701: OK, that makes sense - it's there if they ask for it, but not there by default. So they must be aware. Good.

jgallagher59701 (approved):
Thanks.

PR #594, opened 2025-11-18
DMRPP parser to Pydap dataset model

Mikejmnez:
The following Pull Request:

- [ ] Closes an already existing opened issue #xxxx for fixing a bug, requesting
a new feature, etc.
on a local environment.

### DIFF
--- a/src/pydap/handlers/dap.py
+++ b/src/pydap/handlers/dap.py
@@ -92,7 +92,7 @@ class DAPHandler(BaseHandler):
             self.protocol = "dap4"
             if url.startswith("http://") or url.startswith("https://"):
                 r = session.get(url, stream=True)
-                dmrpp = io.BytesIO(r.content.decode())
+                dmrpp = io.BytesIO(r.content)
                 dmrpp_instance = DMRPPParser(root=ET.parse(dmrpp).getroot())
             else:
                 dmrpp = open(url).read()


src/pydap/handlers/dap.py
```diff
@@ -85,41 +86,60 @@ def __init__(
...
+            self.projection = []
+            self.dataset = dmrpp_instance.to_dataset()
+            self.base_url = dmrpp_instance.data_filepath
+            if not self.base_url:
+                # some dmrpps do not have the opendap url embedded in their
+                if url.endswith(".dap.dmrpp"):
+                    self.base_url = url.removesuffix(".dap.dmrpp")
+                else:
+                    self.base_url = url.removesuffix(".dmrpp")
+        else:
```
jgallagher59701: One thing to know is that Hyrax in NASA's EDC supports DMR++ retrieval using the suffix '.dap.dmrpp'. That should return the DMR++ XML document and I think it performs the URL template resolution, too. Yes. Here's an example: https://opendap.earthdata.nasa.gov/collections/C1996546500-GHRC_DAAC/granules/f16_ssmis_20250221v7.nc.dap.dmrpp

Mikejmnez: About the URL template resolution, you mean for example the: 
```python
xml:base="http://test.opendap.org/opendap/dap4/TestGroupData.nc4.h5"
```
element in the dmrpp? I have seen some `dmrpp`s without this `xml:base` element. And so this conditional is a last resource for finding the opendap url to the file. With earthaccess + virtualizarr, this filepath is provided, but this enables a few more cases (when it is not provided).

jgallagher59701: OK. I remember we had some issues with the xml:base URL in the past. But for the 'Template URL' I meant the data URL the replaces the template code that appears in the XML Dataset element's 'href' attribute. When the documents are written by the code that ingests data in the NASA cloud, the URL to the data granule is not set as te value of that attribute. Instead there is some string like "OPeNDAP_Data_URL" or something like that and the server performs a text substitution before the data access code parses and interprets the DMR++.

PR #619, opened 2026-01-14
bump up dmrVersion parsing to discern two deserialization approaches

### LINKED ISSUE #616: Server can have dmrVersion=1 or dmrVersion=2 Enable support for both
Hyrax can serialize hierarchical data in two ways. It specifies the two versions with `dmrVersion=1` or `dmrVersion=2`. Pydap can support both forms of deserialization, but currently it only knows `dmrVersion=1`. 

## Solution

Must add a conditional (extra logic) for evaluation the dmrVersion to enable the two ways of deserializing data. This has to only apply to Hyrax data servers. TDS already deserializes data in the "correct way" to there will be necessarily a need to further identify the server that produces the response.

Mikejmnez:
The following Pull Request:

- [x] Closes #616. It assumes TDS has specific TDS attributes not present in a Hyrax-produced DMR.
on a local environment.

### TDS
The tds serialization is in accordance to the dap4 spec, and as such, `dmrVersion="1.0"` in the TDS is equivalent to the `dmrVersion="2.0"` from Hyrax. As such, this PR takes the dmrVersion from a TDS dmr and replaces it with a `dmrVersion="2.0"`. However, discerning between a TDS and Hyrax dmr can be tricky. This PR makes the assumption that a TDS-produced dmr always constains the following Attributes:

* _NCProperties
* _DAP4_Little_Endian

I have reached out to Sean from Unidata to confirm this. If so, this PR should be ready to merge.

### Example

*  `dmr_version=1.0` (Hyrax):
The dataset is serialized in the order shown in the tree below (top to bottom, depth-first):
```python
.SimpleGroup.nc4.h5
├──SimpleGroup
│  ├──Y
│  ├──X
│  ├──Temperature
│  └──Salinity
├──time
├──Z
├──Pressure
└──time_bnds
```
The above tree implies that pydap will deserialize `SimpleGroup/Y` first, and `time_bnds` last. This is how Hyrax serializes data (dmrVersion="1.0") but this approach is not in agreement with the way data should be serialized/de-serialized according to the dap4 specification.

*  `dmr_version=2.0`  (Hyrax)
The dataset is serialized in the order shown in the tree below (top to bottom, depth-first)

```python
.SimpleGroup.nc4
├──Pressure
├──time_bnds
├──time
├──Z
└──SimpleGroup
   ├──Temperature
   ├──Salinity
   ├──Y
   └──X
```
In this scenario, `Pressure` is de-serialized first, and `SimpleGroup/X` last. This is the correct way to deserialize, and in accordance to the dap4 spec.

### DIFF
--- a/src/pydap/parsers/dmr.py
+++ b/src/pydap/parsers/dmr.py
@@ -405,10 +405,10 @@ class DMRParser(object):
 
         AttsNames = [subnode.get("name") for subnode in self.node.findall("Attribute")]
         # identify any TDS specific attribute and if so, bump version to 2
-        if all(
+        if any(
             map(
                 lambda x: x in AttsNames,
-                ["_NCProperties", "_DAP4_Little_Endian"],
+                ["_DAP4_Little_Endian", "_dap4.ce"],
             )
         ):
             self.dmrVersion = "2.0"


src/pydap/parsers/dmr.py
```diff
@@ -382,14 +396,40 @@ def __init__(self, dmr):
...
+        # Hyrax has bumped the dmrVersion from 1.0 to 2.0 to indicate proper
+        # deserialization of DAP4 datasets in accordance with the DAP4 spec.
+        # Pydap only supports version 1.0 for now, and so this attribute
+        # will be used to discern between the two versions.
+        if hasattr(self.node, "attrib") and "dmrVersion" in self.node.attrib:
+            self.dmrVersion = self.node.attrib["dmrVersion"]
+
+        AttsNames = [subnode.get("name") for subnode in self.node.findall("Attribute")]
+        # identify any TDS specific attribute and if so, bump version to 2
+        if any(
```
Mikejmnez: it is possible that instead of an "any" conditional, that it is required to have an "all" conditional. However in that scenario, it would exclude the "_dap4.ce" attribute since that is only added when there is a constraint expression

PR #627, opened 2026-01-27
Static logo

aaron-kaplan:
Getting my feet wet in contributing to pydap. Are pull requests welcome?

In the default configuration, where templates are loaded from `pydap.wsgi.templates`, the logo in the upper left of the page is broken:

image

The embedded image link pointed to a file on github, even though logo.png is present in this repo under templates/static. I tried changing that link to point to the local static asset, but that failed at first because StaticMiddleware was only set up to handle text files, not binary/image files. So I'm including a fix for that.

New appearance:
image

To demo: `python -m pydap.wsgi.app`, then visit localhost:8001 in a browser.

### DIFF
--- a/src/pydap/wsgi/app.py
+++ b/src/pydap/wsgi/app.py
@@ -237,9 +237,7 @@ class StaticMiddleware(object):
         content_type, content_encoding = mimetypes.guess_type(full_path)
         package += (".").join([""] + parts)
         if content_encoding is None:
-            body = (
-                importlib.resources.files(package) / resource
-            ).read_bytes()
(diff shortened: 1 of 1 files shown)


aaron-kaplan:
Thanks for the quick review.

I'm disappointed to hear that the server code has been abandoned. I saw recent activity on the repo, and didn't think to check whether any of it was in the server part.

I came to the pydap server after a few frustrating weeks failing to get Hyrax to do what I wanted. Pydap was a breath of fresh air--much simpler and more hackable. I understand that it's probably not up to handling the volume of traffic that Hyrax can handle. We'll see if it's sufficient for our needs.

PR #672, opened 2026-04-08
Fix DODS response parsing to tolerate \r\n line endings

mylesmc123:
Fix DODS response parsing to tolerate \r\n line endings

The Data separator split in safe_dds_and_data, client.py, and dump()
used a strict b"\nData:\n" split which fails against servers that send
b"\nData:\r\n" (e.g. opendap-protocol). Use a find-based approach that
tries both separators without corrupting the binary data payload.

### DIFF
--- a/src/pydap/client.py
+++ b/src/pydap/client.py
@@ -627,8 +627,8 @@ def open_dods_url(
     """Open a `.dods` response directly, returning a dataset."""
 
     r = GET(url, application, session, timeout=timeout)
-
-    dds, data = r.body.split(b"\nData:\n", 1)
+    from pydap.handlers.dap import _split_dds_data
+    dds, data = _split_dds_data(r.body)
     dds = dds.decode(r.content_encoding or "ascii")
     dataset = dds_to_dataset(dds)
     stream = StreamReader(BytesIO(data))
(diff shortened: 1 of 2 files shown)


Mikejmnez:
Thanks @mylesmc123 . DODS is a DAP2 response, which IMHO should be considered legacy. Have you tried with DAP4 on your particular example? For example:
```python
open_url(url, protocol='dap4')
# or
xr.open_dataset(url.replace("https", "dap4")) # some hyrax produce http instead of https. In that case replace http with dap4
``` 

DAP4 should support DAP2 types (see for example [dap4vdap2](https://pydap.github.io/pydap/en/faqs/dap2_vs_dap4.html))

mylesmc123:
Thank you @Mikejmnez. Unfortunately, I couldn't get my particular application working with DAP4. I'm using the [xpublish_opendap plugin](https://github.com/xpublish-community/xpublish-opendap) which uses [opendap-protocol](https://github.com/xpublish-community/opendap-protocol). The opendap-protocol library only implements DAP2 from my understanding and is not being actively maintained. So, I had to use the source code update in my PR to get the xarray call working using an example like:

`ds = xr.open_dataset("dap2://localhost:8080/datasets/somedataset/opendap", engine="pydap")`

PR #688, opened 2026-05-29
Improve data handling from tabular sequences

### LINKED ISSUE #680: Enable creating pyds dataset from DMR containing sequences only
This is the first step towards enabling dap4 sequences. 
## what pydap does

When parsing the metadata pydap flattens the access to any basetype (structures included). This is a safe approach that works for structures but not for sequences. See example below:
```python
url = "http://test.opendap.org/opendap/hyrax/data/ff/gsodock1.dat" # this is a sequence
open_url(url, protocol="dap4").tree()
>>>.gsodock1.dat
├──URI_GSO-Dock.Time
├──URI_GSO-Dock.Depth
├──URI_GSO-Dock.Sea_Temp
├──URI_GSO-Dock.Salinity
├──URI_GSO-Dock.DO 
├──URI_GSO-Dock.pH
├──URI_GSO-Dock.Turbidity
├──URI_GSO-Dock.Air_Temp
├──URI_GSO-Dock.Wind_Speed
├──URI_GSO-Dock.Wind_Direction
└──URI_GSO-Dock.Barometric_Pres

```

Right now, flattening the access is a safe approach. Pydap currently walks the dataset to find the BaseTypes and parses them (Time is a BaseType above for example). That would conflict with `URI_GSO-Dock` which is the parent `Sequence`.

## what can pydap do?

To enable parsing a dmr containing sequences in a safe way that is backwards compatible, pydap should:

1. Check if there is a Sequence somewhere in the dataset.
2. If no Sequence, parse all basetypes in the dmr as currently done.
3. If there is at least one Sequence, parse only the Sequences, Ignoring any other basetype.

Perhaps an additional check to do is:

 if there is a BaseType at the same hierarchy as a Sequence, either fail to continue or flatten any access to the sequence (with a proper warning). Perhaps a suggestion to drop the sequence using a constraint expression.

## End result of this?

- Correctly parse DMR with Sequences.
- This initial approach would mean Sequences cannot coexist with BaseType on their own. For example the following would not be allowed:

```
.gsodock1.dat
├──Variable
└──URI_GSO-Dock
   ├──Time
   ├──Depth
   ├──Sea_Temp
   ├──Salinity
   ├──DO 
   ├──pH
   ├──Turbidity
   ├──Air_Temp
   ├──Wind_Speed
   ├──Wind_Direction
   └──Barometric_Pres
```

Where Variable is a BaseType.

### LINKED ISSUE #692: enable pandas as an optional install to server tabular data
This is one of many initial improvements to pydap server. In using pandas, pydap will no longer need to have a handler per data format. Instead, all the data handling will be done by pandas and a single handler will be needed - > that of pandas.

- This needs to be an optional installation (most people only need the client API)
- Will enable reading of excel, csv, and parquet data files. This will require some additional extra dependencies for each of excel and parquet 
- Because these will be optional / extra imports, these will not be added to the ci/environment.yml file and will not be part of a conda distribution. Thus, imports and testing need to be done with care so that it does not break pydap installation if these extra libraries are not present in the env

Mikejmnez:
The following Pull Request:

- [x] Close #692 : Adds pandas as a tabular store reader. Any tabular data compatible with pandas will be able to be served, as opposed to having to write a data handler per file format. Pandas will do it for us
- [x] Closes #680 (this is already partially addressed by a previous PR). 
- [x] Tests are added to demonstrate a) pydap dataset is correctly generated from a csv/excel file, b) metadata json side car is used in the background to populate attribute metadata, and c) a html representation of the dmr is generated when running pydap as a server.

There are some minor documentation fixes on this PR as well

### example data
```python

Z = -np.array([0, 10, 25, 50, 100, 200, 500, 1000])
T = np.array([22.5, 21.8, 19.2, 15.1, 10.4, 7.2, 4.5, 2.8])
Salt = np.array([34.2, 34.25, 34.4, 34.5, 34.7, 34.9, 34.5, 34.6])

# 1. Create the tabular dataset
data = {
    'depth_m': -Z,
    'temperature_c': T,
    'salinity_psu': Salt
}

df = pd.DataFrame(data, index=None)
df.rename_axis('record', inplace=True)

# 2. View the first few rows
df.head()
# 3. Write into files
df.to_csv(path+file_csv)
>>>

 	    depth_m 	temperature_c 	salinity_psu
record 			
0 	        0 	      22.5 	    34.20
1 	        10 	      21.8 	    34.25
2 	        25 	      19.2 	    34.40
3 	        50 	      15.1 	    34.50
4 	        100       10.4 	    34.70
```

### using pydap
The following uses pandas

```python
from pydap.handlers.tabular_handler import SequenceHandler

file_csv = "data.csv"
path = "<path_to_data>"

pyds1 = SequenceHandler(path+file_csv).dataset
pyds1.tree()
.data1.csv
└──sequence
   ├──record
   ├──depth_m
   ├──temperature_c
   └──salinity_psu

```

### current state

- It produces a DMR correctly 
- One can add a `*.json` sidecar file, that contains metadata attributes. The name of the sidecar **must** match that of the file.
- It knows dap2 filtering/projections when using the pydap data model. See below:
```python
new = pyds1['sequence'][('depth_m', 'temperature_c')][pyds1['sequence.depth_m']<20]
new
>>> .sequence
├──depth_m
└──temperature_c

np.array([data for data in new])
>>> array([[ 0. , 22.5],
       [10. , 21.8]])
```

In the example below, the correct data is returned: temperature and depth data, for depth<20. This is in accordance to the dap2 protocol that pydap implements (in the background) with sequences.


Mikejmnez:
@jgallagher59701 this is ready for review

src/pydap/handlers/tabular_handler/__init__.py
```diff
@@ -0,0 +1,394 @@
+"""A pandas-backed pydap handler for tabular sequence datasets."""
+
+import copy
+import json
+import os
+import re
+import time
+from collections.abc import Mapping
+from email.utils import formatdate
```
jgallagher59701: There might be a smaller package to get access to an ISO 8601 or RFC822 date formatter. This package could have security issues that popup later. This might be me just having come from a meeting about one security issue, so take it with a grain of salt... 

But, if there is another python package that just deals with date formatting... ;-)

src/pydap/handlers/tabular_handler/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+    ".xlsb": ("read_excel", {}),
+    ".ods": ("read_excel", {}),
+    ".odf": ("read_excel", {}),
+    ".json": ("read_json", {}),
+    ".jsonl": ("read_json", {"lines": True}),
+    ".ndjson": ("read_json", {"lines": True}),
+    ".html": ("read_html", {}),
+    ".htm": ("read_html", {}),
+    ".xml": ("read_xml", {}),
+    ".parquet": ("read_parquet", {}),
```
jgallagher59701, line 29: Wow, so we _could_ use this to serve data in parquet! Cool.

src/pydap/handlers/tabular_handler/__init__.py
```diff
@@ -0,0 +1,394 @@
...
+
+import numpy as np
+
+from pydap.exceptions import OpenFileError
+from pydap.handlers.lib import BaseHandler, IterData
+from pydap.model import BaseType, DatasetType, SequenceType
+
+_METADATA_KEYS = frozenset(("dataset", "sequence", "columns"))
+_COMPRESSION_EXTENSIONS = frozenset((".bz2", ".gz", ".xz", ".zip", ".zst"))
+_PANDAS_READERS = {
```
jgallagher59701, line 20: General comment - I think we should limit this to just the formats we test. Keeping this in as a comment is fine, but writing tests for all these formats seems like a huge job. For now, I'd comment out all the formats we're not testing. You could likely use Codex to take a .csv file and make a handful of equivalent files in formats like 'tsv', 'xls', ...

jgallagher59701 (changes requested):

src/pydap/handlers/tabular_handler/__init__.py
Mikejmnez: good catch - `datetime` is the better library for this!

src/pydap/handlers/tabular_handler/__init__.py
Mikejmnez: done - I am narrowed the declaration of fileformats readible via pandas, and added explicit testing for these

jgallagher59701: Cool. Thanks.

jgallagher59701 (approved):
Thanks.

src/pydap/tests/test_handlers_sequence.py
```diff
@@ -0,0 +1,311 @@
...
+    assert sequence["label"].dtype == np.dtype("U")
+    assert list(sequence) == [(1, "one"), (None, None), (3, "three")]
+
+
+def test_supported_tabular_formats_have_dependency_rules():
+    assert set(_FORMAT_DEPENDENCIES) == set(_PANDAS_READERS)
+
+
+@pytest.mark.parametrize("extension", sorted(_PANDAS_READERS))
+def test_supported_tabular_formats_become_single_sequence_dataset(
```
Mikejmnez, line 155: @jgallagher59701 this new test checks that, after writing a file into a tmp directory, pydap using pandas as a data handler can read the file and produce a pydap dataset object.

This test is safe in the sense that, if the required libraries (e.g. fastparquet, etc) used by pandas as "engine" to write/read the files are missing, it skips them. This is important for making conda releases but not so much for ci/cd since pydap is install in server mode always via pip

PR #704, opened 2026-06-25
Upgrade Webob dependency to version 1.8.10 to fix CVE

Scott5S:
This PR adds a minimum required version pin to `Webob` in pyproject.toml to exclude versions of tornado that contain CVE-2026-44889

To reproduce the audit, after cloning and installing the requirements run
```
uv audit --preview --locked
Resolved 98 packages in 3ms
Found 7 known vulnerabilities and no adverse project statuses in 97 packages

Vulnerabilities:

bleach 6.3.0 has 3 known vulnerabilities:

- GHSA-8rfp-98v4-mmr6: Bleach: URI sanitization allows disallowed URI schemes with Unicode > U+00A0 in output

  Fixed in: 6.4.0

  Advisory information: https://github.com/mozilla/bleach/security/advisories/GHSA-8rfp-98v4-mmr6

- GHSA-gj48-438w-jh9v: Bleach clean() / Cleaner() fails to sanitize dangerous URI schemes in allowed formaction attributes

  Fixed in: 6.4.0

  Advisory information: https://github.com/mozilla/bleach/security/advisories/GHSA-gj48-438w-jh9v

- GHSA-g75f-g53v-794x: Bleach linkify(parse_email=True) CPU exhaustion via unbounded email regex scanning

  No fix versions available

  Advisory information: https://github.com/mozilla/bleach/security/advisories/GHSA-g75f-g53v-794x

cryptography 48.0.0 has 1 known vulnerability:

- GHSA-537c-gmf6-5ccf: Vulnerable OpenSSL included in cryptography wheels

  Fixed in: 48.0.1

  Advisory information: https://github.com/pyca/cryptography/security/advisories/GHSA-537c-gmf6-5ccf

tornado 6.5.6 has 1 known vulnerability:

- GHSA-pw6j-qg29-8w7f: Tornado: CurlAsyncHTTPClient leaks per-request credentials on handle reuse

  Fixed in: 6.5.7

  Advisory information: https://github.com/tornadoweb/tornado/security/advisories/GHSA-pw6j-qg29-8w7f

ujson 5.12.1 has 1 known vulnerability:

- GHSA-3j69-69wj-xqx2: UltraJSON: Malformed/Truncated UTF-8 Accepted and Silently Rewritten in ujson.dumps()

  Fixed in: 5.13.0

  Advisory information: https://github.com/ultrajson/ultrajson/security/advisories/GHSA-3j69-69wj-xqx2

webob 1.8.9 has 1 known vulnerability:

- GHSA-fh3h-vg37-cc95: WebOb: Location header normalization during redirect leads to open redirect - again

  Fixed in: 1.8.10

  Advisory information: https://github.com/Pylons/webob/security/advisories/GHSA-fh3h-vg37-cc95
```

This PR is to fix the bottom CVE https://github.com/Pylons/webob/security/advisories/GHSA-fh3h-vg37-cc95

N.B. To install the requirements after pulling, I had to modify the `pyproject.toml` replacing `requires-python = ">=3.11"` with `requires-python = ">=3.12"`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   "requests",
   "requests-cache",
   "scipy",
-  "Webob",
+  "webob>=1.8.10",
   "beautifulsoup4",
   "lxml",
   "Jinja2"


Mikejmnez:
Thanks @Scott5S!

> N.B. To install the requirements after pulling, I had to modify the pyproject.toml replacing requires-python = ">=3.11" with requires-python = ">=3.12"

What do you mean by this? Did the change in this PR made you do this change or is this something you noticed before your change? 
I looked at webob and it should still support py311, and pydap ci/cd includes py311 (for now at least)

Scott5S:
My pleasure @Mikejmnez 
> Did the change in this PR made you do this change or is this something you noticed before your change?

No, I am running the latest [astral/uv](https://github.com/astral-sh/uv/releases/tag/0.11.24) as an environment manager.
On a fresh clone from main (before this commit was merged), I ran

```
uv sync --exclude-newer-package virtualizarr=false

Using CPython 3.13.12
Creating virtual environment at: .venv
  × No solution found when resolving dependencies for split (markers: python_full_version == '3.11.*' and platform_machine == 'ARM64' and sys_platform == 'win32'):
  ╰─▶ Because the requested Python version (>=3.11) does not satisfy Python>=3.12 and virtualizarr>=2.6.0 depends on Python>=3.12, we can conclude that virtualizarr>=2.6.0 cannot be used.
      And because only the following versions of virtualizarr are available:
          virtualizarr<=2.6.0
          virtualizarr==2.6.1
          virtualizarr==2.6.2
          virtualizarr==2.7.0
      we can conclude that virtualizarr>=2.6.0 cannot be used.
      And because pydap[virtualizarr] depends on virtualizarr>=2.6 and your project requires pydap[virtualizarr], we can conclude that your project's requirements are unsatisfiable.

hint: While the active Python version is 3.13, the resolution failed for other Python versions supported by your project. Consider limiting your project's supported Python versions using `requires-python`.
hint: The `requires-python` value (>=3.11) includes Python versions that are not supported by your dependencies (e.g., virtualizarr>=2.6.0 only supports >=3.12). Consider using a more restrictive `requires-python` value (like >=3.12).
```

Note I use `--exclude-newer-package virtualizarr=false` to disable package version exclusions for `virtualizarr`, I have a minimum age limit (to reduce chances of installing compromised versions of dependencies). Without it I would get a failure but for a different reason:
```
hint: `virtualizarr` was filtered by `exclude-newer` to only include packages uploaded before 2026-05-29T08:14:38.742855Z. The latest version satisfying the requirement is v2.7.0, published at 2026-06-25T18:32:21.862Z. Consider using `exclude-newer-package` to override the cutoff for this package.
```

I followed the hint `e.g., virtualizarr>=2.6.0 only supports >=3.12` - so I updated `requires-python` in the `pyproject.toml` and got the `uv sync` to work

Mikejmnez:
Thanks again @Scott5S - I was aware virtualizarr requirements, but I did not realize it would result in such failures since it is an extra dependency. Funny that pip does not raise such a flag but conda might (will make a release soon). Thanks for the heads up!

